PR #1, opened 2017-01-31
[WEEX-17] Abstract a common weex var for each JS framework

Jinjiang:
Accepted weex variable from JS runtime before `legacyFramework.createInstance`.


edwardyoon:
Just FYI, usually one commit per pull request[1] is recommended. Once it synced with apache git, you can't change or clean up commit histories. 

1. http://ndlib.github.io/practices/one-commit-per-pull-request/

PR #21, opened 2017-03-02
WXSlider add event [scrollstart&scrollend]

homeblog:
WXSlider add Android event [scrollstart&scrollend] @


homeblog:
http://velocity.alibaba-inc.com/issues/3346
slider  add onscrollstart onscrollend

PR #23, opened 2017-03-03
* [android] fix the ConcurrentModificationException

misakuo:
The exception:
    

Is due to one thread to modify the ArrayList while another thread is iterating over it, replacing the `ArrayList` with `CopyToWriteArrayList` should be able to solve it


misakuo:
@sospartan review please

sospartan:
This data should only be access by JS thread, u should figure that out.

misakuo:
looks like the inspector's problem, close this pr

PR #25, opened 2017-03-03
Android bugfix wxembed

lybeen:
clear listener when activity destroy to prevent memory leak


sospartan:
Pls dot send PR to master. We don't do that. Use 0.11-dev or dev instead.

PR #28, opened 2017-03-03
* [ios] update sliderComponent , support to setIndex.

boboning:


bluebird78999:
please test this case:http://dotwe.org/weex/7807be5f5e55272340e8d39b6a3db5ed

PR #32, opened 2017-03-03
* [ios] bugfix textarea overflow and rows update

kfeagle:
fixbug:
when text overflow ,text will show
textarea can not set rows and update rows

testcase:http://dotwe.org/weex/13c5ffccdd26c346f106f7d3535e84ad


ios/sdk/WeexSDK/Sources/Component/WXEditComponent.m
```diff
@@ -264,6 +272,10 @@ -(void)setFont:(UIFont *)font
 {
 }
 
+-(void)setRows:(NSUInteger)rows
```
kfeagle: will overwrite by textarea

ios/sdk/WeexSDK/Sources/Component/WXEditComponent.m
```diff
@@ -342,6 +354,13 @@ - (void)updateAttributes:(NSDictionary *)attributes
         _returnKeyType = [WXConvert UIReturnKeyType:attributes[@"returnKeyType"]];
         [self setReturnKeyType:_returnKeyType];
     }
+    if (attributes[@"rows"]) {
```
kfeagle: update row

PR #33, opened 2017-03-03
Fix pos fixed failed when parent is scroller

sospartan:
Fix item with 'position:fixed' scroll with parent.
The reason of this bug is 'fixed' view will add to root node's view. So when the root is 'scroller','fixed' not working.
Testcase: http://dotwe.org/weex/a281d7db58ba73ef0a143535d3703bc4


sospartan:
@zshshr

PR #56, opened 2017-03-07
* [android] ignore system action on WXThread.secure()

misakuo:
Ignore the Runnable send from system


misakuo:
@sospartan review please

PR #60, opened 2017-03-07
* [example] fix bugs for 'npm run build:examples' (vue@2.2.1)

yyqqing:


sospartan:
Do not send to master, use `dev` instead.

PR #69, opened 2017-03-08
[android] fix ConcurrentModificationException in registeration.

sospartan:
Fix fail process when re-register component/module/service, when possible cause ConcurrentModificationException.

@zshshr @littleseven


android/sdk/src/main/java/com/taobao/weex/bridge/WXBridgeManager.java
```diff
@@ -1214,16 +1231,16 @@ public void execJSService(final String service) {
     post(new Runnable() {
       @Override
       public void run() {
-        invokeExecJSService(service);
+        invokeExecJSService(service,mRegisterServiceFailList);
```
littleseven: need a space after ","

PR #71, opened 2017-03-09
* [doc] Update GitHub link, change from alibaba/weex to apache/weex.

DoranYun:
* [doc] Update GitHub link, change from alibaba/weex to apache/weex.


DoranYun:
@Jinjiang

PR #74, opened 2017-03-09
* [android] update modify linear-gradline default value

xkli:
@sospartan


android/sdk/src/main/java/com/taobao/weex/ui/component/WXComponent.java
```diff
@@ -1103,7 +1103,10 @@ public void setBackgroundColor(String color) {
   }
 
   public void setBackgroundImage(String bgImage) {
-    if (!TextUtils.isEmpty(bgImage) && mHost != null) {
+
+    if ("".equals(bgImage) && mHost != null) {
+      getOrCreateBorder().setImage(null);
+    } else if (null != bgImage && mHost != null) {
```
sospartan: what if `bgImage` is string with multiple space?

xkli: Add code to determine if there are multiple spaces

PR #83, opened 2017-03-10
* [ios] 0.10.2 bugfix

acton393:
fix default handler override
remove dlsym for jscore garbage Collection


gunerguner:
`[[UIApplication sharedApplication] openURL:[NSURL URLWithString:@"taobao://"]]; `
Such code will lead to a scheme jump to external taobao application. Could you explain its usage? :(

PR #98, opened 2017-03-13
* doc update new weex-toolkit image

JackPu:
Update weex-project.io web pages image url.
We released a new version of weex-toolkit, some commands have changed. 
So we need replace the old image to new.


Jinjiang:
Your branch is not existed?

PR #101, opened 2017-03-14
0.11 dev bugfix

zshshr:
1. protect NPE crash of support-v7 
2. fix  RuntimeException


android/sdk/src/main/java/com/taobao/weex/ui/component/WXComponent.java
```diff
@@ -1070,7 +1071,11 @@ public final void removeAllEvent() {
     }
     if(mHost != null) {
       mHost.setOnFocusChangeListener(null);
-      mHost.setOnClickListener(null);
+      if(mHost instanceof AdapterView){
```
sospartan: move this detect to list ot whatever.

PR #108, opened 2017-03-15
update doc example about component rect

acton393:
1. update doc example about component rect
2. remove unused zip file


Jinjiang:
Why a file named `WeexDemo.app.zip` added?

acton393:
The `WeexDemo.app.zip`  is for ci test before, now it is useless, so remove it.

Jinjiang:
Squash merged (in order to avoid the large file in the repo & history).
Thanks

PR #118, opened 2017-03-16
[android] post render-only action direct to UI thread.

sospartan:
the 'action' part is not release yet, so modify these public method would not make more 'break change'.

@xkli @YorkShen


YorkShen:
Merge it, pls. I will refactor animation using Action. @sospartan

sospartan:
@YorkShen done.

PR #126, opened 2017-03-17
Jsfm feature vanilla

yong-chen04:
Previous conversations http://mail-archives.apache.org/mod_mbox/incubator-weex-dev/201703.mbox/browser

### DIFF
--- a/html5/frameworks/vanilla/index.js
+++ b/html5/frameworks/vanilla/index.js
@@ -141,6 +141,11 @@ export function getRoot (instanceId) {
  */
 export function receiveTasks (instanceId, tasks) {
   const instance = instances[instanceId]
+
+  // break if parameters is invalid
+  if (instance == null) return
+  if (!Array.isArray(tasks)) return
+
   const { callbacks, document } = instance
   tasks.forEach(task => {
     // `fireEvent` case: find the event target and fire.
--- a/html5/frameworks/vanilla/index.js
+++ b/html5/frameworks/vanilla/index.js
@@ -91,9 +91,12 @@ export function createInstance (
   // It will declare some instance variables like HTML5 Timer APIs etc.
   const instanceVars = Object.assign({
     weex: weexInstanceVar,
-    // deprecated
-    __weex_require_module__: weexInstanceVar.requireModule // eslint-disable-line
+    document // alias for `weex.document`
   }, timerAPIs)
+
+  // wrap IFFE and use strict mode
+  appCode = `(function(global){"use strict";\n ${appCode} \n})(Object.create(this))`
+
   callFunction(instanceVars, appCode)
 
   // Send `createFinish` signal to native.
--- a/html5/frameworks/vanilla/index.js
+++ b/html5/frameworks/vanilla/index.js
@@ -178,7 +178,8 @@ export function registerModules (newModules) {
     newModules[name].forEach(method => {
       if (typeof method === 'string') {
         modules[name][method] = true
-      } else {
+      }
+      else {
         modules[name][method.name] = method.args
       }
     })
@@ -197,7 +198,8 @@ export function registerComponents (newComponents) {
       }
       if (typeof component === 'string') {
         components[component] = true
-      } else if (typeof component === 'object' && typeof component.type === 'string') {
+      }
+      else if (typeof component === 'object' && typeof component.type === 'string') {
         components[component.type] = component
       }
     })
--- a/html5/frameworks/vanilla/index.js
+++ b/html5/frameworks/vanilla/index.js
@@ -129,7 +129,7 @@ export function refreshInstance (instanceId, data) {
  */
 export function getRoot (instanceId) {
   const instance = instances[instanceId]
-  return instance.document.$el.toJSON()
+  return instance.document.body.toJSON()
 }
 
 /**
--- a/html5/frameworks/vanilla/index.js
+++ b/html5/frameworks/vanilla/index.js
@@ -1,107 +1,334 @@
-const config = {}
 
-const instanceMap = {}
+const instances = {}
+const modules = {}
+const components = {}
 
-function init (cfg) {
-  config.Document = cfg.Document
-  config.Element = cfg.Element
-  config.Comment = cfg.Comment
-  config.sendTasks = cfg.sendTasks
+const renderer = {
+  instances,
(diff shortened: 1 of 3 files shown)


html5/frameworks/vanilla/index.js
```diff
@@ -1,107 +1,334 @@
...
+  Object.freeze(weexInstanceVar)
 
-  const result = new Function(...globalKeys)
-  return result(...globalValues)
+  // The function which create a closure the JS Bundle will run in.
+  // It will declare some instance variables like HTML5 Timer APIs etc.
+  const instanceVars = Object.assign({
+    weex: weexInstanceVar,
+    // deprecated
+    __weex_require_module__: weexInstanceVar.requireModule // eslint-disable-line
```
Hanks10100: I think the `__weex_require_module__` should be deleted as it's already deprecated. 

And you can pass `document` here, such as `document: weex.document`, then the developer can use `document` directly instead of `weex.document`.

html5/frameworks/vanilla/index.js
```diff
@@ -1,107 +1,334 @@
...
+  }, timerAPIs)
+  callFunction(instanceVars, appCode)
```
Hanks10100: Since all the js bundle shared the same Weex runtime, I think it's better to wrap an IFFE to the `appCode` before `callFunction`.

See https://github.com/alibaba/weex/pull/1626 for more information.

```js
// wrap IFFE and use strict mode
appCode = `(function(global){"use strict";\n ${appCode} \n})(Object.create(this))`

callFunction(instanceVars, appCode)
```

jiyinyiyong: I can see the benefit of making an object for `global` scope. Any reason why Vue doesn't have one?

html5/frameworks/vanilla/index.js
jiyinyiyong: Better have one in when people are using it frequently. Will do in Monday, the code is in my laptop at office.

jiyinyiyong:
@Hanks10100 any thoughts on the tests? Can I remove `document.addCallback` from tests?

Hanks10100:
The `document.addCallback` is no longer supported, those test cases should be rewritten someday.

html5/frameworks/vanilla/index.js
Hanks10100: Work in progress: https://github.com/vuejs/vue/pull/5049

yong-chen04:
I still got some errors after running the test, do you know how to fix this one(appeared 3 times):

```
  273 passing (751ms)
  61 pending
  3 failing

  1) vanilla test create & destroy instance:
     TypeError: Function.prototype.apply was called on undefined, which is a undefined and not a function
      at Object.sendTasks (html5/runtime/config.js:9:19)
      at Object.createInstance (html5/frameworks/vanilla/index.js:103:12)
      at Context.it (html5/test/unit/vanilla/index.js:35:13)

  2) vanilla test getRoot:
     TypeError: Function.prototype.apply was called on undefined, which is a undefined and not a function
      at Object.sendTasks (html5/runtime/config.js:9:19)
      at Object.createInstance (html5/frameworks/vanilla/index.js:103:12)
      at Context.it (html5/test/unit/vanilla/index.js:55:13)

  3) vanilla test receiveTasks "before all" hook for "getRoot":
     TypeError: Function.prototype.apply was called on undefined, which is a undefined and not a function
      at Object.sendTasks (html5/runtime/config.js:9:19)
      at Object.createInstance (html5/frameworks/vanilla/index.js:103:12)
      at Context.before (html5/test/unit/vanilla/index.js:94:15)
```

Hanks10100:
That because the node.js environment doesn't have the `callNative` method. You should mock that method in tests.

1. Import `sinon`, which is already in the `devDependencies` of package.json.

```js
import sinon from 'sinon'
```

2. Spy the `callNative` API.

```js

describe('vanilla test', () => {
  // In case there did has callNative in the environment
  const originalCallNative = global.callNative

  const callNativeSpy = sinon.spy()
  before(() => {
    global.callNative = callNativeSpy
  })

  afterEach(() => {
    callNativeSpy.reset()
  })

  after(() => {
    global.callNative = originalCallNative
  })

  it('...', ()  => {})
})
```

Moreover, you can use `callNativeSpy` to check the parameter of call native tasks.

yong-chen04:
Sorry I didn't check the notifications earlier.

With sinon now there's only one failure:

```
  1) vanilla test receiveTasks invalid tasks:
     TypeError: Cannot match against 'undefined' or 'null'.
      at Object.receiveTasks (html5/frameworks/vanilla/index.js:144:35)
      at Context.it (html5/test/unit/vanilla/index.js:133:15)
```

Reading by the name of the test, was it designed to fail?

```js
    it('invalid tasks', () => {
      vanilla.receiveTasks('invalid id', [])
      vanilla.receiveTasks(id, {})
      vanilla.receiveTasks(id, [{
        method: 'whatever',
        args: []
      }])
    })
```

yong-chen04:
Also I changed `document.$el` to `document.body` in the last commit to fix a failure:

```diff
-  return instance.document.$el.toJSON()
+  return instance.document.body.toJSON()
```

guessing `document.$el` is no longer available in vanilla framework...

Hanks10100:
As I said in the code review comments, the `receiveTasks` method should check the arguments before using them. However, you can skip the test case of `invalid tasks` either.

yong-chen04:
Since it's the only failure.. Anything else I need to do on this PR?

Hanks10100:
@yong-chen04 The code style may also need to change to follow our eslint rules.

The line 181 and 200 should be: 

```js
if () {
}
else {
}
```

not

```js
if () {
} else {
}
```

yong-chen04:
Updated.

Well, I thought you still writing `else` in that old way...
https://github.com/apache/incubator-weex/blob/master/html5/render/browser/extend/components/scrollable/scroll.js#L68
https://github.com/apache/incubator-weex/blob/master/examples/vue/showcase/include/util.js#L59
https://github.com/apache/incubator-weex/blob/dev/build/webpack.test.config.js#L17

jiyinyiyong:
Any plan to merge this update?

jiyinyiyong:
@fkysly ping?

PR #132, opened 2017-03-20
[jsfm] * fix freeze error when assign name to Error

yuanyan:
make below code work:
```
var error = new Error(); 
error.name = 'Invariant Violation';
```


yuanyan:
Merged

PR #138, opened 2017-03-21
[doc] use inline event handler binding.

MrRaindrop:
fix doc examples: recommend to use inline function binding to deprecate the access of `e.target.attr`.


Jinjiang:
Is there any reason to put the doc update & package.json update together? It's better to divide them as two independent commits.
Thanks.

MrRaindrop:
Not necessarily needed. package.json updates removed.

PR #141, opened 2017-03-21
Class  common's problem 

nicefan:
I use `<style src="assets/common.styl" lang="stylus"/>`  require class, want class shared，but no good, becasue I view the build dist file, these class all repeat, is not one module pack in.


sospartan:
We will merge 0.11-dev to master when v0.11 released. So this PR will not proceed currently. Thank you.

PR #143, opened 2017-03-22
* [android] slider improvement

misakuo:
Changes:
- allows auto play when the number of items is 2
- allows infinite scroll when the number of items items is 2
- ability to disable infinite scroll

Testcase:
- http://dotwe.org/weex/f7e8b0cb0e0a2bfe613011b28d8c0276

Known issue:
- When the number of items is 2,  the loop scroll from item 2 to item 1 will have no animation effect


sospartan:
@misakuo why not create some cases about slider 😄

misakuo:
macaca let's go!

PR #155, opened 2017-03-24
* [android] input & textarea improvement

misakuo:
Changes:
- support `type="number"`
- fire the `keyboardevent` event when the softkeyboard is toggled or hided

Demo:
- http://dotwe.org/vue/df1cc1d6d2920f966c8d222b38dc44aa


sospartan:
The code seems fine to me. But the event is called 'keyboardevent' literally?  The '-event' look like unnecessary.

PR #173, opened 2017-03-27
Update README.md

JonasGao:
It seems that '0' does not work, so I replaced '1'


JonasGao:
This is the list use '0'
===



This is the list use '1'
===

PR #175, opened 2017-03-28
* [ios] fix embed components frame when resizing root view frame

Neeeo:
embed component frame isn’t changed when resizing root view frame


acton393:
@Neeeo  thx for your contributions

PR #190, opened 2017-03-30
【iOS】if setFrame (0, 0, width, 0), the height will adjust with content, but setFrame again, it fail ,the height become zero.

czcomt:


cxfeng1:
please use english commit log

cxfeng1:
please use english commit log and create another pull request.

PR #191, opened 2017-03-30
* [android] prevent the NullPointerException

misakuo:
prevent the NPE that caused by WXSDKInstance object is null


android/sdk/src/main/java/com/taobao/weex/WXSDKManager.java
```diff
@@ -218,7 +218,11 @@ public static WXSDKManager getInstance() {
   }
 
   public static int getInstanceViewPortWidth(String instanceId){
-    return getInstance().getSDKInstance(instanceId).getInstanceViewPortWidth();
+    WXSDKInstance instance = getInstance().getSDKInstance(instanceId);
+    if (instance == null) {
+      return 750;
```
sospartan: Use a constant instead.

PR #195, opened 2017-03-30
fix method onActivityResult bug,resultCode is error passed

lltemplar:


sospartan:
pls create PR based on same branch.

PR #203, opened 2017-03-31
add link from customer view to actual dom.

wispy316:
Android customer view has no link to the actual dom. so there is no way to access from the View to virtual dom ( we can call WXComponet.getDomObj() on the actual dom to get virtual dom).

Added _public T getComponent();_  to _IRenderStatus_ , everything shines.


android/sdk/src/main/java/com/taobao/weex/ui/view/IRenderStatus.java
```diff
@@ -211,4 +211,6 @@
 public interface IRenderStatus<T extends WXComponent> {
 
   public void holdComponent(T component);
+
+  public T getComponent();
```
zshshr: 1. do not modify interface directly ,check downward  compatibility problem

wispy316:
Change it to IRenderResult.

wispy316:
travis-ci always fails. change to #229

PR #215, opened 2017-04-06
* [html5] scrollToElement method support no animation

fkysly:
scrollToElement method support no animation


fkysly:
this could be closed, refer to #226

PR #217, opened 2017-04-06
[android] disable click event when gesture already consume touch event

sospartan:
Disable click event when gesture is consumed same serial touch-event.

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/view/gesture/WXGesture.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/view/gesture/WXGesture.java
@@ -256,6 +256,7 @@ public class WXGesture extends GestureDetector.SimpleOnGestureListener implement
   private WXGestureType mPendingPan = null;//event type to notify when action_up or action_cancel
   private int mParentOrientation =-1;
   private boolean mIsPreventMoveEvent = false;
+  private boolean mIsTouchEventConsumed = false; //Reset to false when first touch event, set to true when gesture event fired.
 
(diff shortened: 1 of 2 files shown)


android/sdk/src/main/java/com/taobao/weex/ui/view/gesture/WXGesture.java
```diff
@@ -309,12 +320,14 @@ public boolean onTouch(View v, MotionEvent event) {
...
           break;
         case MotionEvent.ACTION_UP:
         case MotionEvent.ACTION_POINTER_UP:
           finishDisallowInterceptTouchEvent(v);
           result |= handleMotionEvent(LowLevelGesture.ACTION_UP, event);
           result |= handlePanMotionEvent(event);
+          mIsTouchEventConsumed = mIsTouchEventConsumed || result;
           break;
         case MotionEvent.ACTION_CANCEL:
           finishDisallowInterceptTouchEvent(v);
```
YorkShen: Reset flag to false when it is canceled?

sospartan: It's not necessary to reset flag here,`Pan` may already trigger by 'move' .

YorkShen: Reset to false when **cancel** fired?

sospartan: click event should not trigger when any of gesture events (start/move)  fired. So `ACTION_CANCEL` will not reset flag.

PR #219, opened 2017-04-06
chang viewport adapting algorithm.

MrRaindrop:
# intro

Request to enable viewport adapting algorithm changing for weex-vue-render. The previous way to adapt different screens is by setting viewport meta in the head. The compatibility in some devices is un-tested and the implementation for this meta in some browsers is not so reliable. What's more, some user-defined components in this changed viewport environment is not working correctly, since those components assume current viewport is a more common value, like device-width, and they draw UI depending on this.

So this pr solved this problem by introducing realtime recalculation for component's styles, which are scaled-up/down value to adapt current device-width, to enable the consistency in all the devices with varies of screens. This pr also solve the problem that the scale values used in  `dom.scrollToElement` and `animation.transition` wouldn't adapt to different screens.

This pr also enable runtime autoprefix for styles, which means users don't have to put `$processStyle` or `autoprefixer` in their vue-loader config. The render will do this for users automatically and they don't have to worry about it.

# changelog:

* change viewport adapting algorithm by recalculating styles during runtime.
* discard $processStyle and autoprefixer config for vue-loader.
* add test cases.


dongnaebi:
in `<style>` element,88px will computed to 44px in iPhone6.but inline-style dose not work,and import style like `<style src="../style/main.css"></style>` dose not work too.

MrRaindrop:
@dongnaebi can you give me a simple demo to show this error with inline style ? I tested in a simple case and got no error found, maybe you have something not mentioned in this case to get this style computing error ? BTW, import src is not valid to use in weex, since it's not a scoped style. You can only use `<style scoped>` tag in weex.

dongnaebi:
@MrRaindrop sorry,弄反了，is `<style>` tag dose not computed.
it's not an error,is inconsistent performance.

I was test it in npm package v0.11.5:
http://dotwe.org/vue/cf99c75bbadfa61c4ae151b4c430ffb4

dongnaebi:
here somthing I found.
my viewport is `<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">`
if change it to 0.5,it's work fine,but `weex.config.env.deviceWidth` will be 1500,some calculations based on deviceWidth will be large.

MrRaindrop:
This branch has changed viewport adapting algorithm. So the npm version is out dated (BTW, the latest version on npm is `0.11.2`) when this branch is merged, and the device-width will be the viewport size consistently (e.g. 375 for iphone6).

@dongnaebi  In your case I suggest you don't change your viewport. If you want to use a scale based on your design width, such as 750px, you can use `<meta name="weex-viewport" content="750">`. This would be ok.

You can specify your viewport width to 750px with `<meta name="weex-viewport" content="750">`. This meta works for all the `0.11.x` version of weex-vue-render, what is changed is the adapting algorithm. You shouldn't set the viewport meta as `device-width` and `xx-scale=1` since the `env.deviceWidth` and `env.deviceHeight` will be calculated relying on this. What you can change is the `weex-viewport` meta.

dongnaebi:
OK,thanks,good man

PR #229, opened 2017-04-07
add link from customer view to actual dom. copy from #203.

wispy316:
Android customer view has no link to the actual dom. so there is no way to access from the View to virtual dom ( we can call WXComponet.getDomObj() on the actual dom to get virtual dom).

Added public T getComponent(); to IRenderStatus , everything shines.

#203 always failed at travis with unkown reason. (failed at iOS and html5, but no relative change in this pr)
just create a branch from dev-0.12, then try again.


wispy316:
@zshshr  @sospartan   dear, have a look at this Pull request. 

Ignore the failure of travis CI.  
I have tried with empty modification, travis CI failed too. #247

PR #235, opened 2017-04-09
[ios]Explicit start samples by terminal.

CodeLife2012:
Open use  the default system application to open the file, maybe user has  changed  to a different application, then  start sample will fail.


CodeLife2012:
Don't known  why `Copy Pods Resources `  is deleted, I just change the last line of `Start Samples`.

cxfeng1:
Seems good to me,  @acton393  Please have a code review

acton393:
@CodeLife2012  cool,  thx for your contribution, that's good ~~~

PR #238, opened 2017-04-10
* [ios]  change keyboard event to keybord and add animation for Navig…

kfeagle:
* [ios]  change keyboard event to keybord and add animation for navigationBarHidden


ios/sdk/WeexSDK/Sources/Controller/WXBaseViewController.m
```diff
@@ -70,10 +70,10 @@ - (void)viewDidLoad
     self.view.backgroundColor = [UIColor whiteColor];
     self.automaticallyAdjustsScrollViewInsets = NO;
     [self _renderWithURL:_sourceURL];
-    
     if ([self.navigationController isKindOfClass:[WXRootViewController class]]) {
-        self.navigationController.navigationBarHidden = YES;
```
acton393: @boboning   animating ?

PR #271, opened 2017-04-18
【iOS】 Make Weex Support DP

yuedongweng:
because Android Support DP, so iOS need support too。


cxfeng1:
@yuedongweng  the PR is not right for your purpose.

PR #278, opened 2017-04-21
[android] fix viewport conflict. One is static value store in WXSDKIn…

wispy316:
fix viewport conflict. One is static value store in WXSDKInstance, an other is instance value. The dom object use instance value, when static value is not equal to instance value, the wx unit will be smaller than normal.


android/sdk/src/main/java/com/taobao/weex/dom/WXStyle.java
```diff
@@ -284,10 +311,6 @@ public float getBorderRadius() {
     return temp;
   }
 
-  public float getBorderTopWidth() {
-    return getBorderWidth(Constants.Name.BORDER_TOP_WIDTH);
-  }
-
```
wispy316: move a few lines down, together with getBordedXXXWidth()

android/sdk/src/main/java/com/taobao/weex/WXSDKInstance.java
```diff
@@ -185,18 +184,6 @@ public boolean isNeedValidate() {
...
-  /*
-  *  store custom ViewPort Width
-  */
-  @Deprecated
-  public void setViewPortWidth(int viewPortWidth) {
-    mViewPortWidth = viewPortWidth;
-  }
-
-  @Deprecated
-  public static int getViewPortWidth() {
```
zshshr: do not delete the api

PR #298, opened 2017-04-27
+ [html5] add jsonpCallback support

elover:
stream 模块支持jsonpcallback自定义，因为有的api会先经过cdn缓存，为了避免jsonpcallback name穿透缓存，所以希望可以添加jsonpcallback的的自定义功能。

### DIFF
--- a/html5/render/browser/extend/api/stream.js
+++ b/html5/render/browser/extend/api/stream.js
@@ -12,7 +12,7 @@ let jsonpCnt = 0
 const ERROR_STATE = -1
 
 function _jsonp (config, callback, progressCallback) {
-  const cbName = 'jsonp_' + (++jsonpCnt)
+  const cbName = config.jsonpCallback || 'jsonp_' + (++jsonpCnt)
   let url
 
   if (!config.url) {


html5/render/browser/extend/api/stream.js
```diff
@@ -12,7 +12,7 @@ let jsonpCnt = 0
 const ERROR_STATE = -1
 
 function _jsonp (config, callback, progressCallback) {
-  const cbName = 'jsonp_' + (++jsonpCnt)
+  const cbName = config.jsonpCallback || 'jsonp_' + (++jsonpCnt)
```
MrRaindrop: I think it's better to use `jsonpCallbackName` instead of `jsonpCallback`, since there'are already two parameters about callback function, `callback` and `progressCallback`, in this case another `xxCallback` may mislead user to pass a jsonp callback **function** other than the callback name **string** to it according to the other `xxCallback`s are all **functions**. Using `jsonpCallbackName` should avoid this misunderstanding.

gurisxie:
HI, @elover , Thanks a lot for your contribution about weex. I have merged your [pr-717](https://github.com/apache/incubator-weex/pull/717) ,please close this pr.

Hope you continue to take part in weex.
Best wishes!

PR #319, opened 2017-05-05
* [test] change border test case to fit android

cxfeng1:


cxfeng1:
@YorkShen  Please have a review.

PR #327, opened 2017-05-08
* [jsfm] update jsfm to 0.20.7

atomtong:
update jsfm to 0.20.7


atomtong:
jsfm support Event bubbleing and broacast  in Vue framework

PR #342, opened 2017-05-14
+ [ios] ios support prerender

kfeagle:
+ [ios] ios support prerender


MappleHero:
Can you give us a example that how to use this feature?

PR #35, opened 2017-03-06
Support waterfall and recycler

cxfeng1:
* Add recycler component 
* support  multi-column layout
* [ios] support delete header in the middle

PR #349, opened 2017-05-16
* [android] fixed issues #35. when a tag href is empty

IskenHuang:
As https://issues.apache.org/jira/browse/WEEX-35 ，Let iOS and Android A component open empty url the same logic

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/component/WXA.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/component/WXA.java
@@ -50,7 +50,7 @@ public class WXA extends WXDiv {
         ImmutableDomObject domObject = getDomObject();
         if (domObject != null) {
           WXAttr attr = domObject.getAttrs();
-          if (attr !=null && (href = (String)attr.get("href")) != null) {
+          if (attr != null && (href = (String)attr.get("href")) != null && !TextUtils.isEmpty(href)) {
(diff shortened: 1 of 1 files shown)


acton393:
iOS  is #352

sospartan:
Already fixed by https://github.com/apache/incubator-weex/commit/eeaf303aecc091d37d8b6d9b16d88dd5e5c97ed2    
Sorry, @IskenHuang your PR is deprecated 🐶

IskenHuang:
@sospartan deprecated is ok. When href is empty string jump to current page logic the same with web. But web support href execute javascript. For example `<a href="javascript:void(0);"></a>`. In this case when click on a component will to do nothing on web. But on iOS/Android will open a new page 'javascript:void(0);'. It would be 404 page. When we want to get the same effect should be to do as below sample:

```
<a v-if="isGotoNewPage" :href="myURL"><text>New page</text></a>
<a v-if="!isGotoNewPage"><text>New page</text></a>

// or
<div @click="myAction()"><text></text></div>

export default {
    methods: {
        myAction() {
            if(isGotoNewPage) weex.requireModule('navigator').push(this.myURL)
        }
    }
}
```

The a component's purpose is reduce js/native interactive times. But this changes it's not enough for real case.

sospartan:
@IskenHuang execute js with `href` is not possible in native currently.
I suggest you to discuss about this in dev@ mail list.

PR #352, opened 2017-05-16
* [ios] use bundlejs url when the href for A component is empty

acton393:


acton393:
android case is  #349

PR #387, opened 2017-05-24
[iOS] add x-cache info 

xujk2008:


ios/sdk/WeexSDK/Sources/Model/WXSDKInstance.m
```diff
@@ -259,6 +259,11 @@ - (void)_renderWithRequest:(WXResourceRequest *)request options:(NSDictionary *)
             }
             return ;
         }
+        
+        if([response isKindOfClass:[NSHTTPURLResponse class]]){
+            NSString* cacheType = [[(NSHTTPURLResponse*)response allHeaderFields] valueForKey:@"X-Cache"];
+            strongSelf.userInfo[@"X-Cache"] = cacheType;
```
acton393: the userInfo maybe nil here, maybe should new an dictionary

PR #390, opened 2017-05-25
+ [ios] update maraca case

kfeagle:
+ [ios] update maraca case


kfeagle:
https://travis-ci.org/apache/incubator-weex/builds/235909073?utm_source=github_status&utm_medium=notification

ci pass

PR #395, opened 2017-05-25
[android]  js exception adapter change to list. so that can handler m…

wispy316:
[android] js exception adapter change to list. so that can handler more than one listener.   

Now faced a problem is 
WeexErrorCallback (used by Motu weex js exception caught) 
**is conflict with** 
JSExceptionGlobalEventReport (added in taobao_weex_adapter to caught js exception and throw it to js, we use JSTracker to log exceptions.)


wispy316:
do it in WeexErrorCallBack at crashreportadapterLoader

PR #405, opened 2017-05-27
add jslibVersion and SdkVersion perf point for perfAnalysis

atomtong:
add jslibVersion and SdkVersion perf point for perfAnalysis


acton393:
@atomtong  please check the danger comments

PR #438, opened 2017-06-10
h5-render 2.0 add more features.

MrRaindrop:
Hi, I add some more features to h5-render 2.0, including:
* support `position: sticky` in list / scroller's direct children.
* use `resetLoadmore` method to enable list / scroller to be able to trigger loadmore event again.
* set alignment for children of loading / refresh to `center` by default in both x and y axis.
* add `waterfall` component.

and also fix several bugs:
* fix triggering duplicate appear / disappear event.
* hide scroll bar by default.


MrRaindrop:
jira: https://issues.apache.org/jira/browse/WEEX-46

PR #441, opened 2017-06-12
* [android] remove some reference of DomObject on input

misakuo:
JIRA Issue: https://issues.apache.org/jira/browse/WEEX-41


sospartan:
0.13?

misakuo:
Will we continue to release the next version from this branch? This is an urgent bugfix

PR #442, opened 2017-06-12
* [ios] cache attributedString for improve performance while draw Text

acton393:


sospartan:
Pls create a JIRA issue about this. And notice the dangerjs comments.

PR #447, opened 2017-06-14
+ [ios] not input \n when return key type  is not set default

kfeagle:
+ [ios] not input \n when return key type  is not set default


sospartan:
Create a JIRA issue about this, pls.

PR #448, opened 2017-06-15
* [android] enable gesture for slider

misakuo:
I am already testing the basic gesture and it working well


misakuo:
I will update the document in another PR

PR #459, opened 2017-06-21
Android bugfix repeat adding event

cairuiquan:
The animation.vue sample, all botton's click event will call twice


cairuiquan:
The animation.vue sample，all botton click event will invoke twice

PR #468, opened 2017-06-25
Ios bugfix mamaca 0.14 dev 0624

kfeagle:
Ios bugfix mamaca


test/scripts/components/textarea-maxlength.test.js
```diff
@@ -43,11 +43,13 @@ describe('@ignore-android textarea maxlength vue test2 ', function () {
     return driver
       .elementById('textarea')
       .sendKeys('12345678')
-      .sleep(2000)
+      .sleep(3000)
+      .source()
```
kfeagle: this will display all element . user can gain the message

PR #471, opened 2017-06-26
[example] Modify the countdown example

Hanks10100:
Rename `countdown` to `count-down`, because "countdown" is reserved tag name in weex-vue-framework.


elover:
why not change reserved tag to remove "countdown"?

Hanks10100:
@elover I'll do that too. Its code is in the vuejs/vue repo.

PR #476, opened 2017-06-27
[android] Upgrade css-layout to yoga library

sospartan:
Related issue : https://issues.apache.org/jira/browse/WEEX-34
Upgrade `css-layout` to `yoga`.

This is a huge PR, better reviewing by commits.
## Notable changes:
WXDomObject now inherit YogaNode, which is wrapper for object in jni.
minSDKVersion change to 15, the yoga library required. I’ve make a PR to yoga, and it accepted by yoga. So yoga will change minSDKVersion to 14. After that, we can change it back to 14 too.
Spacing from old css-layout library will stay, because the ImmutableDomObject use it.
Clean css-layout code will in another seperate commit.
Some new rule:
- 'Only leaf nodes with custom measure functionsshould manually mark themselves as dirty
- 'Cannot add child: Nodes with measure functions cannot have children'
Measure Function in Weex are modified according these rules.
The ‘clone’ method from WXDomObject is deprecated. Use a ‘asResult’ to return the data object need by components.


netputer:
Any progress?

gurisxie:
weex will not use yoga,so close this pr .thanks @sospartan

PR #488, opened 2017-07-03
* [android] change 'contentSize' value in list 'scroll' event, use value

sospartan:
caculate by layout engine. @not-danger
https://issues.apache.org/jira/browse/WEEX-55


misakuo:
great job! 
can `scroller` component do that too?

sospartan:
@misakuo scroller may not have this issue, the size should stable in `scroller`.

PR #489, opened 2017-07-04
[doc] Add documents for "weex.supports" API

Hanks10100:
The `weex.supports` API will be added in v0.15, and only supported in Weex DSL 2.0 (.vue).


sospartan:
ping @Tancy

MrRaindrop:
weex-vue-render should implement this too, right ? Otherwise it'll be crashed in web platform.

PR #495, opened 2017-07-05
+ [ios] fix bug if duration is 0 or negative, animation time is given 0.25

kfeagle:
+ [ios] fix bug if duration is 0 or negative, animation time is given 0.25

example
http://dotwe.org/vue/4ac17cab8845cd4a5fd469e930ea9b9e

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Module/WXAnimationModule.m
+++ b/ios/sdk/WeexSDK/Sources/Module/WXAnimationModule.m
@@ -361,7 +361,7 @@ WX_EXPORT_METHOD(@selector(transition:args:callback:))
         layer.frame = originFrame;
     }
     
-    if(WXFloatEqual(animation.duration, 0) || WXFloatLessThan(animation.duration, 0)){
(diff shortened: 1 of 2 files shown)


ios/sdk/WeexSDK/Sources/Module/WXAnimationModule.m
```diff
@@ -355,7 +360,14 @@ - (void)_createCAAnimation:(WXAnimationInfo *)info
         layer.anchorPoint = CGPointZero;
         layer.frame = originFrame;
     }
-    [layer addAnimation:animation forKey:info.propertyName];
+    
+    if(WXFloatEqual(animation.duration, 0) || WXFloatLessThan(animation.duration, 0)){
```
acton393: may be we can provide more function like lessOrEqualThan  and  and greatOrEqualThan value, here will be more less code

kfeagle: I use !WXFloatGreaterThan. it only one judgement and no need add method :-)

PR #501, opened 2017-07-10
[iOS] fix WXCycleSliderComponent initial index

JiaHongXu:
bug fix: initial index of WXCycleSliderComponent may be incorrect if
inserting subviews after layoutDidFinish (occurs when creating
time-consuming subview/components)


ios/sdk/WeexSDK/Sources/Component/WXCycleSliderComponent.m
```diff
@@ -460,6 +460,12 @@ - (void)insertSubview:(WXComponent *)subcomponent atIndex:(NSInteger)index
                 }
             }
             [recycleSliderView insertItemView:view atIndex:index - offset];
+            
+            // check if should apply current contentOffset
+            // in case inserting subviews after layoutDidFinish
+            if (index-offset == _index) {
```
kfeagle: _index > 0 is it better?

PR #507, opened 2017-07-13
[vue-render] support fast click & env.osName & hairlines border.

MrRaindrop:
* support fast click
* use `weex.config.env.osName` to tell is Android or iOS.
* support hairlines (1px device width) border.


sospartan:
@Hanks10100 review this PR.

PR #550, opened 2017-07-28
[jsfm] Fix potential memory leak in callbacks

Hanks10100:
+ [x] pass instance id to callback manager.
+ [x] Destroy `taskCenter` when destory the `document`.
+ [x] Change the type of `callbacks` from array to object.
+ [x] Support to return values in `fireEvent`.
+ [x] Support to send `options` as last argument in `callNativeModule` and `callNativeComponent`.


Hanks10100:
It's already merged in #559 .

PR #571, opened 2017-08-03
[iOS]Of the 32-bit device, BOOL's encode is' c ', and the lack of this code will lead to errors.

bb-coder:


acton393:
hi @bb-coder   did you try it on 32bit device and there is an error ?

PR #572, opened 2017-08-04
modify WXValidateProcessor to make  options params avalible@notdanger

woshiaxie:
modify WXValidateProcessor and WXBridgeManager.etc to make  options params avalible.


android/sdk/src/main/java/com/taobao/weex/dom/WXDomObject.java
```diff
@@ -627,7 +627,7 @@ public String dumpDomTree() {
    * @param json the original JSONObject
    * @return Dom Object corresponding to the JSONObject.
    */
-  public static  @Nullable WXDomObject parse(JSONObject json, WXSDKInstance wxsdkInstance){
```
zshshr: do not modify public method

PR #609, opened 2017-08-17
*[android]fix the bug of extra data not add to param in image load event @notdanger

fengcunhan:


misakuo:
Is the iOS have the `extra` field now?

fengcunhan:
I do not check extra field in iOS.

Hanks10100:
The CI is broken and the PR have no response for a long time. I'll close it shortly. If there still need further discussions about it, you could open a new pull request.

Thanks for your cooperation.

PR #622, opened 2017-08-21
* [test] bugfix caused by mocha retries changes 

gurisxie:
* [test] bugfix caused by mocha retries changes
1、test script delete some case
2、test script logic change


test/pages/components/switch-event.vue
```diff
@@ -11,11 +11,12 @@
     <wxc-desc>
       <text class='desc'>
 测试点：
```
acton393: highly recommend to be English

PR #678, opened 2017-09-05
+ [ios] add missing scroller delegate impl

hjhcn:
we have a requirement   **add delegate to WXScrollerComponent** , but now can only get `scrollViewDidScroll ` callback , so that we cannot impl like this

```

- (void)initScroller {
//scroller is  a instance of WXScrollerComponent
    self.scroller = (id<WXScrollerProtocol>)self.source;
    [self.scroller addScrollDelegate:self];
}

- (void)remove*** {
    [self.scroller removeScrollDelegate:self];
}

...

- (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView {
    [self doSomeThing:@"start"];
}
- (void)scrollViewDidEndDecelerating:(UIScrollView *)scrollView {
    [self doSomeThing:@"end"];
}

...





```

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Component/WXScrollerComponent.m
+++ b/ios/sdk/WeexSDK/Sources/Component/WXScrollerComponent.m
@@ -509,6 +509,71 @@ WX_EXPORT_METHOD(@selector(resetLoadmore))
     }
 }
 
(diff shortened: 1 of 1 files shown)


gurisxie:
HI, @hjhcn , please add tc file pr with code, that we can better judge correctness

PR #686, opened 2017-09-06
* [ios] fix analyze error case @notdanger

acton393:
fix Objective-C usage analyzed by clang analyze.


gurisxie:
please descripe what you has updated in this pr description.

PR #696, opened 2017-09-13
* [ios] clear toast queue when instance disappear @notdanger

acton393:


ios/sdk/WeexSDK/Sources/Model/WXSDKInstance.m
```diff
@@ -378,7 +379,8 @@ - (void)destroyInstance
     if(url.length > 0){
         [WXPrerenderManager addGlobalTask:url callback:nil];
     }
-    
+    NSMutableArray * toastQueue = [[WXToastManager sharedManager] valueForKey:@"toastQueue"];
```
kfeagle: toastQueue 会不会是空值？

acton393: Singleton method for sharedManager  will initialize toastQueue as a mutableArray， here I just removeAllObject.  In other side , if toastQueue is nil,  sending message to nil value will take no effect.

gurisxie:
has no question

PR #698, opened 2017-09-14
* [ios] support multi roles for accessibiliity

acton393:


gurisxie:
review result,not found anything error

PR #702, opened 2017-09-14
Fix typo on homepage

rubenoost:


rubenoost:
The key of the translation should probably also be changed, but i do not have enough knowledge of the project to determine where this should happen.

PR #715, opened 2017-09-20
Update common-event.md

yurizhang:
## Longpress event

If a `longpress` event is bound to a component, the event will be triggered when user long press on it.    
**Notes: ** The `input` and `switch` component does not currently support the `click` event, please use `change` or `input` event instead.    

It' should't "click", should be "longpress"

### DIFF
--- a/doc/source/references/common-event.md
+++ b/doc/source/references/common-event.md
@@ -24,7 +24,7 @@ The onclick attribute fires on a click gesture on the element.
 ## Longpress event
 
 If a `longpress` event is bound to a component, the event will be triggered when user long press on it.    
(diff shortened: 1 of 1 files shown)


yurizhang:
"The input and switch component does not currently support the click event, please use change or input event instead."

Have a small issue: It' should't "click", should be "longpress"
中文文档也是。

gurisxie:
@yurizhang  we don't suggest not use `click` event on switch, because change or input event can instead of it.

Hanks10100:
The CI is broken and the PR have no response for a long time. I'll close it shortly. If there still need further discussions about it, you could open a new pull request.

Thanks for your cooperation.

PR #717, opened 2017-09-20
* [html5] add custom jsonp callback name support

orifake:
Add custom jsonp callback name support, refer to #298


orifake:
Is there anyone can review this pr?

PR #724, opened 2017-09-21
* [doc] fix doc dotwe address

atomtong:
* fix doc dotwe scheme
* bug issuse link https://issues.apache.org/jira/browse/WEEX-72

### DIFF
--- a/doc/source/guide/set-up-env.md
+++ b/doc/source/guide/set-up-env.md
@@ -90,4 +90,4 @@ You can modify this page in `src/foo.vue`. The code is shown below:
 </style>
 ```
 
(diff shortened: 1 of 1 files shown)


jondong:
Thanks for the quick fix, actually I'm preparing a PR for this, not only for the dotWe issue, but also hope to cover the English expression issue. Anyway, I'll propose another supplementary patch for this afterward.
BTW, reviewing & submitting your own PR is not a good practice. We need to avoid this. From the perspective of regulation, I hope we can make reviewing codes by another committer as a forcing step before submit, even you are a committer with the merging right.

PR #749, opened 2017-09-27
  * [android] fix android new Date() cannot get accuracy time

yuhun-alibaba:
[ISSUE ID] https://aone.alibaba-inc.com/issue/12621307
  if we use var time = new Date(), cannot get accuracy time on some android platform such as nexus5, android5.0 rom.
  the root cause is miss ANDROID_ROOT envp when run jsc process, cause libc init exception


gurisxie:
@yuhun-alibaba please not use alibaba issue that others can't not reach.

also,I guess you need bugfix on version 0.16,please pr to release-0.16 branch.

Thanks.

PR #754, opened 2017-09-28
* [ios]  support attributes to set scrollVertical when the scrollView contentSize smaller than the scrollview frame size 

acton393:
* [ios] support attributes to set scrollVertical when the scrollView contentSize smaller than the scrollview frame size 

try case : http://dotwe.org/vue/7f349e0545703a3f17d7757e27c80849

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Module/WXWebSocketModule.m
+++ b/ios/sdk/WeexSDK/Sources/Module/WXWebSocketModule.m
@@ -126,7 +126,7 @@ WX_EXPORT_METHOD(@selector(onclose:))
 
 - (void)close:(NSString *)code reason:(NSString *)reason
 {
-    if(!code)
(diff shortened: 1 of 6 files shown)


ios/sdk/WeexSDK/Sources/Module/WXWebSocketModule.m
```diff
@@ -126,7 +126,7 @@ - (void)close
 
 - (void)close:(NSString *)code reason:(NSString *)reason
 {
-    if(!code)
+    if([WXUtility isBlankString:code])
```
gurisxie: if code = nil ,has any question?

acton393: there is a assert crash  when code is empty string but not a nil value

PR #757, opened 2017-09-29
* [doc] Fixed grammar glitches in set-up-env.

jondong:
Updated grammar and expression glitches in "Setup Development
Environment", also added dotWe link to the corresponding Chinese
version of this document.

Bug: WEEX-76


jondong:
@cxfeng1 @gurisxie PTAL, thanks.

PR #769, opened 2017-10-09
0.16 dev md5

gubaojian:
1、record md5 when js execute exception
2、fix scroll to element not scroll when none scroll action


gurisxie:
@gubaojian , please pr to release-0.16 , if you need bugfix 0.16 version

PR #770, opened 2017-10-09
[WEEX-78] Replace "includes" with "indexOf" to improve compatibility

Hanks10100:
### Problem

`Array.prototype.includes` is defined in ECMAScript 2016 (ES7) which is not compatible with low version js engines.

### Solution

Use `Array.prototype.indexOf` instead of `Array.prototype.includes`.

### DIFF
--- a/html5/runtime/vdom/Element.js
+++ b/html5/runtime/vdom/Element.js
@@ -409,7 +409,7 @@ export default class Element extends Node {
 
     if (!isStopPropagation
       && isBubble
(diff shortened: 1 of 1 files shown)


Hanks10100:
Fix [WEEX-78](https://issues.apache.org/jira/projects/WEEX/issues/WEEX-78).

Hanks10100:
Already merged. https://github.com/apache/incubator-weex/commit/d10be0bcf9624ca0f56fe8f60d646f71ed2f1aa0

PR #774, opened 2017-10-09
fix scroll to bug and  add md5 for template

gubaojian:
1、record md5 when js execute exception
2、fix scroll to element not scroll when none scroll action

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/utils/WXFileUtils.java
+++ b/android/sdk/src/main/java/com/taobao/weex/utils/WXFileUtils.java
@@ -29,6 +29,10 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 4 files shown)


liweiting:
点个赞

android/sdk/src/main/java/com/taobao/weex/utils/WXFileUtils.java
```diff
@@ -130,4 +134,26 @@ public static boolean saveFile(String path, byte[] content, Context context) {
...
+    } catch (UnsupportedEncodingException e) {
+      return  "";
+    }
+  }
+
+  public static String md5(byte[] bts){
+    try {
+      MessageDigest digest = MessageDigest.getInstance("MD5");
+      digest.update(bts);
+      BigInteger bigInt = new BigInteger(1, digest.digest());
```
YorkShen: Using new String(byte []) is a better choice.

PR #783, opened 2017-10-11
* [ios] bugfix list view can not scroll

acton393:


ios/sdk/WeexSDK/Sources/Component/WXScrollerComponent.m
```diff
@@ -232,13 +243,13 @@ - (void)updateAttributes:(NSDictionary *)attributes
         ((UIScrollView *)self.view).scrollEnabled = _scrollable;
     }
     if (attributes[@"alwaysScrollableHorizontal"]) {
-        _alwaysScrollableHorizontal = [WXConvert BOOL:attributes[@"alwaysScrollableHorizontal"]];
-        ((UIScrollView*)self.view).alwaysBounceHorizontal = _alwaysScrollableHorizontal;
+        _alwaysScrollableHorizontal = [WXConvert NSString:attributes[@"alwaysScrollableHorizontal"]];
```
gurisxie: why has twice transform?

acton393: actually I want to save the string value , as bool can not expression third case.

PR #785, opened 2017-10-11
[WEEX-17] Abstract "weex" variable for each js framework

Hanks10100:
Fix [WEEX-17](https://issues.apache.org/jira/browse/WEEX-17).

Pass the `weex` variable to each front-end framework in the last argument of `createInstance`. Its API is defined [here](http://weex.apache.org/references/weex-variable.html).

Both Vue and Rax can use the `weex` variable with the same implementation. Currently, Vue passed `weex` variable to the js bundle, but Rax doesn't pass it yet.


gurisxie:
I have merged it,but bot seems broken.

Hanks10100:
Got, this PR can be closed.

PR #788, opened 2017-10-12
[Android]bug fix for list cell add event has none affect when cell not init

gubaojian:
1、bug fix for list cell add event has none affect when cell not init
2、sticky offset supported
3、keep position lost bugfix
4、sticky bugfix when none view


gurisxie:
has merged... but bot seems broken.

gubaojian:
good, code has merged

android/sdk/src/main/java/com/taobao/weex/ui/component/list/BasicListComponent.java
```diff
@@ -236,6 +245,9 @@ protected T initComponentHostView(@NonNull Context context) {
     if (transforms != null) {
       bounceRecyclerView.getInnerView().addItemDecoration(RecyclerTransform.parseTransforms(getOrientation(), transforms));
     }
+    if(getDomObject().getAttrs().get(Constants.Name.KEEP_POSITION_LAYOUT_DELAY) != null){
```
misakuo: Need to judge whether it is null?

PR #791, opened 2017-10-13
* [ios] fix issue https://issues.apache.org/jira/browse/WEEX-81  that…

acton393:
fix issue https://issues.apache.org/jira/browse/WEEX-81 that box outer box-shadow is abnormal while using border-radius.

new : image

before:  
image


gurisxie:
has mergerd,but bot seems broken.

PR #796, opened 2017-10-14
0.16 format delete log

yuhun-alibaba:
delete debug log and format log on so


gurisxie:
has merged... but bot seems broken.

PR #801, opened 2017-10-17
* [android] fix bug if application is null log not print out

atomtong:
* [android] fix bug if application is null log not print out


gurisxie:
has rebase,but bot seems broken.

PR #803, opened 2017-10-17
 * [android] fix on android 4.0.* platform cannot startup multiprocess

yuhun-alibaba:
* [android] fix on android 4.0.* platform cannot startup multiprocess  
 since on android 4.1 use ANDROID_PIE to ensure safety
 on android api 15 or below use libweexjst.so to startup new process


gurisxie:
has merged. but bot seems broken.

PR #816, opened 2017-10-23
* [android] memory optimization for box-shadow

misakuo:
Reduce drawing count and memory use for box-shadow.
Now we can reduce **3/4** memory use on creating the box-shadow bitmap.


misakuo:
@gubaojian have a review please

PR #822, opened 2017-10-25
Remove maven central repository, use only jcenter

yinonavraham:
jcenter is a super set on top of maven central, so having both of those repositories is redundant, and the preferred one should be jcenter.

CheckList:

  * [x] Update API docs for the component.
  * [x] Update/Add demo to demonstrate new feature.
  * [x] Update test scripts for the feature.
  * [x] Add unit tests for the feature.

(checked, although not really relevant - not a feature nor bug...)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,6 +1,7 @@
 
 buildscript {
     repositories {
+        mavenLocal()
         jcenter()
     }
     dependencies {
@@ -14,15 +15,18 @@ plugins {
 }
 
 repositories {
+    mavenLocal()
     jcenter()
 }
 
 subprojects {
     repositories {
+        mavenLocal()
         jcenter()
     }
     buildscript {
         repositories {
+            mavenLocal()
             jcenter()
         }
         dependencies {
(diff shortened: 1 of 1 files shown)


misakuo:
Thanks for your contribution.  we can remove the `mavenCentral` from repositories, but I think to replace it with `mavenLocal` will be better?

yinonavraham:
Certainly not. `mavenLocal` is your local maven cache (i.e. the `.m2` directory). Doing so will remove any central maven repository. You should at least have either `jcenter` or `mavenCentral`, I defenetly prefer `jcenter`.

misakuo:
I mean 
```groovy
 repositories {
        mavenLocal()
        jcenter()
  }
```
First find in the local maven cache, if not found, find in jcenter

yinonavraham:
Ah, sure, that's ok, sorry for the misunderstanding. I think it is implicit, but it's ok to have it anyway.

android/build.gradle
```diff
@@ -1,7 +1,6 @@
 
 buildscript {
     repositories {
-        mavenCentral()
```
misakuo: Replace it with `mavenLocal`

PR #830, opened 2017-11-01
[jsfm] Support receive tasks in weex-js-runtime

Hanks10100:
Move the `callJS` and `receiveTasks` into a separate file. This feature has no breaking change.

By support the `receiveTasks` in `weex-js-runtime`, this method can be removed from `weex-vue-framework` and `weex-rax-framework`. It can keep the js-native communication have the same behavior.


MrRaindrop:
+1

PR #834, opened 2017-11-05
* [ios] support max-width when drawing text

acton393:
fix issue: https://issues.apache.org/jira/projects/WEEX/issues/WEEX-91?filter=allopenissues

Here is the code on  dotwe: http://dotwe.org/vue/7e556c582a55630518d0bf6336ddde0e
I test compatibility on iOS 8,iOS9 and iOS 11.


cxfeng1:
It looks good to me,  @gurisxie please accept the pull request.

PR #844, opened 2017-11-08
[WEEX-99][Android] fix setViewport: sometimes doesn't work

myeveryheart:
[WEEX-99][Android] fix setViewport: sometimes doesn't work.

The setViewport method in WXMetaModule is ASYN right now which makes weexInstance.viewportWidth setted after view created.

Bug:99

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/module/WXMetaModule.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/module/WXMetaModule.java
@@ -39,7 +39,7 @@ public class WXMetaModule extends WXModule {
     public static final String WIDTH = "width";
     public static final String DEVICE_WIDTH = "device-width";
 
-    @JSMethod(uiThread = false)
+    @JSMethod
     public void setViewport(String param) {
         if (!TextUtils.isEmpty(param)) {
             try {


cxfeng1:
@gubaojian please have a review.

gubaojian:
@myeveryheart  can you give an failed case on dotwe.org

myeveryheart:
@gubaojian Sorry, dotwe.org don't support entry.js. But you can try this like doc
`
// entry.js
import App from './app.vue'
const meta = weex.requireModule('meta')
// 配置 viewport 的宽度为 640px
meta.setViewport({width: 640})
App.el = '#root'  
new Vue(App)
`
After that, you can see that width sometimes doesn't work.
By the way, it was known that we should set UI on main thread, which ASYN doesn't.

gurisxie:
HI, @myeveryheart there is something wrong with your pr.

It will not fill full when setViewport = 640.

please have a look.

the tc link：[viewport tc](https://github.com/apache/incubator-weex/blob/master/test/pages/modules/viewport-basic.vue)

the result pic:
image

gubaojian:
@myeveryheart   javascript is run on JSThread， uiThread = false， the method will be called on js thread sync. if you changed it on main thread, it will become async

PR #846, opened 2017-11-09
[jsfm] Support register and trigger component hooks

Hanks10100:
Support to receive the `componentHook` task via `callJS`, and expose the `registerHook` and `triggerHook` APIs in the task center.

This feature can be used to achieve the lifecycle hook within `<recycle-list>`. The design proposal can be found in https://github.com/Hanks10100/weex-native-directive/tree/master/component .

The front-end frameworks (Vue.js, Rax) rely on this feature to implement the virtual component.


MrRaindrop:
+1

PR #847, opened 2017-11-09
[jsfm] Add a tracker for legacy ".we" framework

Hanks10100:
Add a tracker to collect the bundle type usage message. If the developer is still using the legacy ".we" framework, it'll print an error in console and record the bundle messages.

References: [WEEX-90: Remove The Legacy Weex DSL 1.0 (.we) Front-End Framework](https://issues.apache.org/jira/browse/WEEX-90)


MrRaindrop:
+1

PR #851, opened 2017-11-13
[WEEX-107] [android] weex transition support on android platform

gubaojian:
[WEEX-107] [android] weex transition support on android platform, improve batch time when layout

1. transition animation supported on android platform

2.improve batch performace in DOMActionContextImpl， improve batch time when layout

3.edit animation module time function get same interceptor with web.


misakuo:
Great function

PR #860, opened 2017-11-14
[WEEX-112][iOS] make weex playground app adapt to iPhone X

acton393:
according to https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/launch-screen/ ,
we must add a 1125px X 2436px  launch image for Portrait size and 2436 X1125px,  at the same time,
developers must consider about the safe-area of container.

Bug: 112

image


gurisxie:
This PR has been merged and will close soon, thanks ^_^

PR #864, opened 2017-11-15
[iOS] Make sure gesture recognizers won't be lost.

AntiMoron:

CheckList:

**if** *isBugFix* **:**

  * [x] Make sure that you add at least one unit test for the bug which you had fixed.

**elif** *isNewFeature* **:**

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Model/WXComponent.m
+++ b/ios/sdk/WeexSDK/Sources/Model/WXComponent.m
@@ -54,7 +54,9 @@
     NSMutableDictionary *_styles;
     NSMutableDictionary *_attributes;
     NSMutableArray *_events;
-    
+    // To make sure view's gestureRecognizers won't be lost.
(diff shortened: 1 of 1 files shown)


AntiMoron:
Here's the issue.

weexsmall

On bounce triggered, gesture recognizers of cells outside screen are lost in the gif.

cxfeng1:
@AntiMoron  Can you provide the demo? http://dotwe.org/vue/

AntiMoron:
@cxfeng1  No,  customized weex component is involved. I can send you the demo.

acton393:
hi @AntiMoron  I highly recommend you add your gesture in `ViewDidLoad` method, which is the component lifecycle,  if the component is child of list, it will be re-used, actually it will be set as nil, and when it come to appear , then `loadView` lifecycle called again, and this tip I will mention it in our documentation later,  so there is no need to reserve view's `gesture recognizer`.

PR #869, opened 2017-11-15
[WEEX-101][iOS] specify voice-over navigation order is column or vertical

acton393:
Consider an app that shows items in vertical columns. Normally, voiceOver would navigate through
these items in horizontal rows. Setting the value of this property to YES on the parent element of the
items in the vertical columns causes VoiceOver to respect the app’s grouping and navigate them.Andthe default of this property is NO.

Bug:101


cxfeng1:
Is there any discussion result for the feature on Android？

PR #882, opened 2017-11-17
[WEEX-127][android]Use HashSet Instead of array, because some app not

gubaojian:
Use HashSet Instead of array, because some app not contains arrayset support library


gurisxie:
has merged.

PR #908, opened 2017-11-23
* [doc] doc example: delete <text> props “lines”

SimplyY:
`<text>` component props “lines” is invalid, so I delete “lines=3” in all doc files.

weex playgroud demo（which proved what I said）: http://dotwe.org/vue/5ac8c3970d049c1c0b41c4ef9a103963

### DIFF
--- a/doc/source/cn/references/components/text.md
+++ b/doc/source/cn/references/components/text.md
@@ -97,10 +97,10 @@ version: 2.1
 <template>
   <div class="wrapper">
     <div class="panel">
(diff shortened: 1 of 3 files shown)


acton393:
hi @SimplyY 
It looks like `lines` property is useless on your case , but I remember it can be used to limit characters into 1 row and ellipse in text.

your case should take into consideration, support lines applied when text drawing

SimplyY:
http://dotwe.org/vue/b1825c31715032e000492290e94a2281

look this.

only 'useCss' class（lines in css） is work, lines props is still not work.

SimplyY:
@acton393

Hanks10100:
Just remind you that, the website and documents have been moved to the [incubator-weex-site](https://github.com/apache/incubator-weex-site) repo. The `doc` folder here has already been removed. Please refer to #993 .

PR #909, opened 2017-11-24
* [doc] doc references: update weex.config.bundleUrl's description

SimplyY:
weex.config.bundleUrl's description is confused. 

I thinked it is jsUrl without queryString ago.I spent a half of day, finally know bundleUrl is  the same as URL of the page.

so I come here to change doc for people can quickly get the answer in future.


acton393:
thx for your contribution @SimplyY , it will be accepted and merged, later on.

PR #913, opened 2017-11-27
[WEEX-109][iOS] bugfix with scrollstart and scrollend event's and doc update

doumafang:
refix some bug about slow drag scoller with no animation to get 'scrollend' event's bug.

And in this pr, we fix scrollend event can not be called's bug, also, add document about scroller's scrollstart and scollend.
test case: http://dotwe.org/vue/ba8045f24bf6e1265bb51e94b5ef418c

Besides, fix transition doc version in this pr.

### DIFF
--- a/doc/source/cn/references/common-style.md
+++ b/doc/source/cn/references/common-style.md
@@ -344,7 +344,7 @@ transform 属性向元素应用 2D 转换。该属性允许我们对元素进行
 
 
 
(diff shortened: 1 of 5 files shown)


cxfeng1:
commit log should confirm to https://github.com/apache/incubator-weex-site/blob/master/source/development-process.md#commit-guidelines

PR #915, opened 2017-11-28
[WEEX-148][Android] box-shadow support multi shadows

misakuo:
Testcase:
http://dotwe.org/weex/f1d7342879a6d1f52ee85c9e810d8df9

Expected effect:    
img


gurisxie:
please check whether it's ok. If yes, you need to change the base png for diff.
image

PR #918, opened 2017-11-28
[WEEX-150] Implement AppShell for iOS.

jondong:
This is the first patch for implementing AppShell for iOS.

Basically, this patch created a minimized appshell to demonstrate
the usage of Weex, the MainViewController contains a WeexSDKInstance
which loads a precompiled javascript page in bundlejs. It also
contains a WXImgLoaderProtocol implementation to facilitate the image
loading.


jondong:
@cxfeng1 @acton393 PTAL, thanks

acton393:
@jondong  It looks good. But I am concerning about how the Weex developers can add their own customization using appShell, maybe documentation is needed here.

jondong:
@acton393 thanks for the feedback, that's right we need documentation, my plan is to implement appshell for both iOS and Android firstly, then announce them in mail list as well as adding documents for them, then remove the playground from Weex SDK at last. I'm wondering if this is the right chance to provide documentation with this PR because the feature is not quite ready yet. Anyway I can add the plan into the umbrella bug WEEX-149 to track the process. Is it sound good to you?

appshell/ios/Podfile
```diff
@@ -0,0 +1,10 @@
+source 'https://github.com/CocoaPods/Specs.git'
+platform :ios, '8.0'
+
+target 'AppShell' do
+    pod 'SDWebImage', '~>4'
+    pod 'SocketRocket', '~>0.5'
+    pod 'WeexSDK', '>=0.15'
```
acton393: here may be the   pod 'WeexSDK', '~>0.15',  so it will get the latest version of SDK

jondong: Thanks for the comment, I've updated the codes.
Anyway '>=0.15' will make sure the version number is no less than 0.15, which also helps the project to get the latest WeexSDK.

jondong:
@acton393 PTAL again, thanks.

PR #926, opened 2017-12-02
* [jsfm] fix the invalid cache in weex.requireModule

Hanks10100:
There is an invalid cache in `weex.requireModule` when refreshing the same instance. This PR fix it.

For example, refresh this http://dotwe.org/vue/229cd943053747a2d483b2530df12d2d page, the `dom.scrollToElement` won't work.


Hanks10100:
This has already been merged in https://github.com/apache/incubator-weex/commit/ec2d234e578652160cf967867946ba82b2bae5be

PR #928, opened 2017-12-02
* [jsfm] add debug utils

Hanks10100:
Move tracker to a separate file and add more debug utils.


Hanks10100:
This has already been merged in https://github.com/apache/incubator-weex/commit/4c89ea93deefe993b1120bd17e95bf49e5eabf06

PR #935, opened 2017-12-07
WEEX-164[iOS] input component: add hideDoneButton attribute to input to hide the done button of keyboard

linkhui:
When we use input component, There is a done button. but sometime we don't want this done button. so we add the hideDoneButton attribute to input.
This is the demo page:
http://dotwe.org/vue/2439876b3a2ec60026bf839b6fbb046b


linkhui:
This is the demo:
http://dotwe.org/vue/2439876b3a2ec60026bf839b6fbb046b

PR #938, opened 2017-12-11
 [WEEX-169] [android] Fix new String null pointer crash when byte array is null

yuhun-alibaba:
Fix new String null pointer crash when byte array is null


gubaojian:
good

PR #941, opened 2017-12-14
[WEEX-172][android] improve jsservice useage，support mutil type params

lybeen:
1. modify register jsservice options type to Object, thus can pass whatever type params
2. modify jsservice script to support js object
3. add method getService


zshshr:
take care of that  Changing the  parameter type of public Api  may cause problem of compatibility

PR #942, opened 2017-12-14
[WEEX-171][WXEnvironment] using another flag to determine whether open devtool or not

Rowandjj:
another flag to allow developers to enable/disable devtool


android/sdk/src/main/java/com/taobao/weex/bridge/WXBridgeManager.java
```diff
@@ -123,6 +123,7 @@
   public static final String KEY_ARGS = "args";
 
   // args
+  public static final boolean sForceEnableDevTool = false;
```
misakuo: Moving it to WXEnvironment.java

Rowandjj:
done

PR #955, opened 2017-12-22
[WEEX-90][example] remove legacy .we examples

Hanks10100:
Remove source codes of .we examples, then rebuild the examples, replace the generated bundles in ios and android.


Hanks10100:
@MrRaindrop @gubaojian @acton393 Please take a review.

MrRaindrop:
+1

PR #960, opened 2017-12-28
[jsfm] Enhance the multi-instance isolation (sandbox)

Hanks10100:
According to the discussion in the mailing list ([Use multi-context for Weex page](https://lists.apache.org/thread.html/87bf83253ed7ee318a8cf98d693bc534b3e08b4d28655a1290f2e0c7@%3Cdev.weex.apache.org%3E)), the current isolation strategy is vulnerable, distinguish the **"Global Context"** and the **"Instance Context"** is a more reasonable solution.

In practice, **the logic of instance management and code execution will be moved from js framework to native render engines.**

This PR implement a new `createInstanceContext` API instead of the former `createInstance` to create the instance context for each page. Refer to [this PR](https://github.com/vuejs/vue/pull/7272) to see the modification in the Vue.js framework. Rax doesn't support it yet.

As for native render engines, they should also call `createInstanceContext` instead of `createInstance`, and parse the bundle type of the code (Vue or Rax) and then send it to js framework.


Hanks10100:
@yacheng @acton393 @yangshengtao Please take a review.

PR #966, opened 2018-01-04
[android][WEEX-160]update style diff, when update, need clone to sync dom object

gubaojian:
Weex Transition Improvement for high frequency , update style diff, when update, need clone to sync dom object


acton393:
anyone who can review it

PR #970, opened 2018-01-05
Fixed link to weex-hackernews

whitelynx:


Hanks10100:
Thanks for your contribution, but all documents of Weex are moved to https://github.com/apache/incubator-weex-site now, and this link is [removed in the new documents](https://github.com/apache/incubator-weex-site/blob/f42e30d80cfb4b29d88cece64c63f148660faf26/source/guide/advanced/use-vuex-and-vue-router.md).

PR #981, opened 2018-01-15
+ [android] support "word-break" style in text

sunsky6:
* Support "word-break" style in text. Can use "keep-all" or "break-all".


sunsky6:
Doc: [word-break](https://developer.mozilla.org/zh-CN/docs/Web/CSS/word-break)

PR #996, opened 2018-01-22
#import <coreText/CoreText.h>   ==>   #import <CoreText/CoreText.h>

zerozhu:
In iOS SDK WXUtility.m file, the #import <coreText/CoreText.h> header file  written error


acton393:
@zerozhu  thanks for the typo, it will be accepted.

PR #1004, opened 2018-01-25
JsService should be registered during weex reload

Darin726:
An app may reload weex sdk in some situation.

So we should register js service during  weex sdk reloading


yuhun-alibaba:
LGTM

PR #1014, opened 2018-02-05
* [android] close #123, websocket operate on main thread

Gentrio:
DefaultWebSocketAdapter will throw an exception when run on android 7.0

[https://developer.android.com/about/versions/nougat/android-7.0-changes.html](url)

> Due to a bug in previous versions of Android, the system did not flag writing to a TCP socket on the main thread as a strict-mode violation. Android 7.0 fixes this bug. Apps that exhibit this behavior now throw an android.os.NetworkOnMainThreadException. Generally, performing network operations on the main thread is a bad idea because these operations usually have a high tail latency that causes ANRs and jank.


misakuo:
Duplicate with #1015

PR #1018, opened 2018-02-07
[Android] Fix box-shadow style will show black color shadow when you use two or more decimal places alpha value

sunsky6:
This is a testcase: [http://dotwe.org/vue/3a80180a163ac983f22a16feaddc60e7](http://dotwe.org/vue/3a80180a163ac983f22a16feaddc60e7).

When our use two or more decimal places alpha value ( example: `box-shadow: 0 0 20px 0 rgba(0, 0, 0, 0.06)` ) in `box-shadow` style.

**The result should like this:**
image

**but now is like this:**
image

**These change fix this bug. Thx.**


misakuo:
Great job, finally, we have ushered in the second person who understands box-shadow logic 😄

PR #1028, opened 2018-02-12
[WEEX-217][iOS] fix:WXTransform should not crash while parsing 'translate(0)' on iOS

dechaos:
According to the latest published W3C specification about CSS Transforms:

translate() = translate( <length-percentage> [, <length-percentage> ]? )
specifies a 2D translation by the vector [tx, ty], where tx is the first translation-value parameter and ty is the optional second translation-value parameter. If <ty> is not provided, ty has zero as a value.

, translate(tx) is equal to translate(tx, 0). In the previous version of Weex, we removed the array length check in method [WXTransform parseTranslate:]. if the parser encounters a 'translate(0)', the array contains only one single value inside, parseTranslate will fetch the element at index 1 from the array, which causes a typical out-of-bounds exception, and lead to an app crash eventually.

We should add the array length check back to the method to avoid the crash in the above case.

As we known in many js packing procedures, 'translate(x, 0)' will be compressed/minified to the form 'translate(x)’, Weex should avoid such inconsistent implementations and conforms the W3C specifications.

Bug: 217

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Component/WXTransform.m
+++ b/ios/sdk/WeexSDK/Sources/Component/WXTransform.m
@@ -349,7 +349,9 @@
 - (void)parseTranslate:(NSArray *)value
 {
     [self parseTranslatex:@[value[0]]];
-    [self parseTranslatey:@[value[1]]];
+    if (value.count > 1) {
+        [self parseTranslatey:@[value[1]]];
+    }
 }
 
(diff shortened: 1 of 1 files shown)


acton393:
@dechaos  thank you for you contribution, and the file change looks good to me,  a very specific explanation at the same time for CSS of W3C rules, let's make it more standard.

PR #1045, opened 2018-03-02
[Android] using android.util.Base64 instead of fastjson.util.Base64

misakuo:


misakuo:
merged at #f77db8fe529fd42b632ef4eb25f70313679de4c5

PR #1047, opened 2018-03-05
[WEEX-233][iOS] More enhanced about <web> component

tw93:
Hi, I'm Tw93(侑夕) from [Fliggy](https://www.fliggy.com/), I'd like to discuss the enhancements about [<Web>](https://weex.apache.org/references/components/web.html) component, let it take over where we can't implement through weex.

## Target

`<web>` is used to display web content that specified by src attribute in weex page,We also can use webview module to control WebView behavior such as goBack, goForward and reload.

But it's not enough in most businesses，I think a more enhanced web component should have the following:

* Support to send messages from Weex to a html in <web> component.
* Support to send messages from a html in <web> component to Weex.
* Support to render html source.

Previously, <web> component was an island that only rendered a remote url, but enabled it to communicate through the context native to the webView. Then I'm going to describe the solution that I came up with.

### Code
* Weex iOS:
  * [WXWebComponent.m](https://github.com/tw93/incubator-weex/blob/ios-feature-enhanced-web/ios/sdk/WeexSDK/Sources/Component/WXWebComponent.m)
  * [WXWebComponent.h](https://github.com/tw93/incubator-weex/blob/ios-feature-enhanced-web/ios/sdk/WeexSDK/Sources/Component/WXWebComponent.h)
  * [WXWebViewModule.m](https://github.com/tw93/incubator-weex/blob/ios-feature-enhanced-web/ios/sdk/WeexSDK/Sources/Module/WXWebViewModule.m)
* Weex Vue Demo:
  * [http://dotwe.org/vue/3576ed477d21dc020e63bf57dc414fff](http://dotwe.org/vue/3576ed477d21dc020e63bf57dc414fff)
  * [Demo Bundle JS](http://dotwe.org/raw/dist/3576ed477d21dc020e63bf57dc414fff.bundle.wx)
  * [Web End](http://h5.m.taobao.com/trip/wx-detection-demo/web.html)

### [Demo](https://gw.alipayobjects.com/zos/rmsportal/sBfTGamezxaBCGPshyXT.gif) && [Theory](https://img.alicdn.com/tfs/TB1hEhXbxGYBuNjy0FnXXX5lpXa-1872-1208.png)

<table><tr><td></td><td></td></tr></table>

#### Support to send messages from Weex to a html in <web> component.

* Native：(W3c [MessageEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent))
 
  ```
   // Weex postMessage to web
  - (void)postMessage:(NSDictionary *)data {
      WXSDKInstance *instance = [WXSDKEngine topInstance];
  
      NSMutableString *bundleUrlOrigin = @"";
  
      if (instance.pageName) {
          NSString *bundleUrl = [instance.scriptURL absoluteString];
          NSURL *url = [NSURL URLWithString:bundleUrl];
          bundleUrlOrigin = [NSString stringWithFormat:@"%@://%@%@", url.scheme, url.host, url.port ? [NSString stringWithFormat:@":%@", url.port] : @""];
      }
  
      NSDictionary *initDic = @{
          @"type" : @"message",
          @"data" : data,
          @"origin" : bundleUrlOrigin
      };
  
      NSString *json = [WXUtility JSONString:initDic];
  
      NSString *code = [NSString stringWithFormat:@"(function (){window.dispatchEvent(new MessageEvent('message', %@));}())", json];
      [_jsContext evaluateScript:code];
  }
  ```
  
* How to use：
 
  ```
    // Weex
    const webview = weex.requireModule('webview');
    // recommend
    this.$refs['wxc-web'].postMessage({ detail: "a message" });
    //not recommend
    //webview.postMessage(this.$refs.webview, {detail:"a message"});
  
    // Web 
    window.addEventListener('message',function(e){
    console.log(e.data,e.type,e.origin); // { detail: "a message" },"message","*"
  },false)
  ```

#### Support to send messages from a html in <web> component to Weex.

- Native：(W3c [MessageEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent))

  ```
  //Weex catch postMessage event from web
      _jsContext[@"postMessage"] = ^() {
  
          NSArray *args = [JSContext currentArguments];
  
          if (args && args.count < 2) {
              return;
          }
  
          NSDictionary *data = [args[0] toDictionary];
          NSString *origin = [args[1] toString];
  
          if (data == nil) {
              return;
          }
  
          NSDictionary *initDic = @{ @"type" : @"message",
                                     @"data" : data,
                                     @"origin" : origin
          };
  
          [weakSelf fireEvent:@"message" params:initDic];
      };
  ```

- How to use：

  ```
    // Weex
    <web @message="onMessage"></web>
    
     onMessage (e) {
            console.log(e.data,e.type,e.origin); // { detail: "a message" },"message","*"
      },
    
    // Web iframe
      window.parent.postMessage({ detail: "a message" }, '*');
  ``` 

#### Support to render html source.

* Native:
 
  ```
  [_webview loadHTMLString:_source baseURL:nil];
  ```
* How to use:
 
  ```html
  <web source='<p style="text-align: center;"><img align="absmiddle" src="http://img03.taobaocdn.com/imgextra/i3/1124701655/TB2zmmIcpXXXXbIXpXXXXXXXXXX_!!1124701655.jpg"/></p>'></web>
  ```

Welcome to put forward any suggestion about the solution or other requirements for the <web> component, Thanks！

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Component/WXWebComponent.m
+++ b/ios/sdk/WeexSDK/Sources/Component/WXWebComponent.m
@@ -49,6 +49,8 @@
 
 @property (nonatomic, strong) NSString *url;
 
+@property (nonatomic, strong) NSString *source;
+
 @property (nonatomic, assign) BOOL startLoadEvent;
 
 @property (nonatomic, assign) BOOL finishLoadEvent;
@@ -69,6 +71,11 @@ WX_EXPORT_METHOD(@selector(goForward))
 {
     if (self = [super initWithRef:ref type:type styles:styles attributes:attributes events:events weexInstance:weexInstance]) {
         self.url = attributes[@"src"];
+        
+        if(attributes[@"source"]){
+            self.source = attributes[@"source"];
+        }
+        
     }
     return self;
 }
@@ -88,14 +95,22 @@ WX_EXPORT_METHOD(@selector(goForward))
     _webview.opaque = NO;
     _jsContext = [_webview valueForKeyPath:@"documentView.webView.mainFrame.javaScriptContext"];
     __weak typeof(self) weakSelf = self;
+
+    // This method will be abandoned slowly.
     _jsContext[@"$notifyWeex"] = ^(JSValue *data) {
         if (weakSelf.notifyEvent) {
             [weakSelf fireEvent:@"notify" params:[data toDictionary]];
         }
     };
     
+    _jsContext[@"postMessage"] = ^(JSValue *data){
+        [weakSelf fireEvent:@"notify" params:[data toDictionary]];
+    };
+
     if (_url) {
         [self loadURL:_url];
+    }else if(_source){
+        [_webview loadHTMLString:_source baseURL:nil];
     }
 }
 
@@ -104,6 +119,11 @@ WX_EXPORT_METHOD(@selector(goForward))
     if (attributes[@"src"]) {
         self.url = attributes[@"src"];
     }
+    
+    if(attributes[@"source"]){
+        self.source=attributes[@"source"];
+    }
+    
 }
 
 - (void)addEvent:(NSString *)eventName
@@ -133,6 +153,21 @@ WX_EXPORT_METHOD(@selector(goForward))
             [self loadURL:_url];
         }
     }
+}
+
+- (void) setSource:(NSString *)source
+{
+    NSString *newSource=[source copy];
+    if(!newSource){
+        return;
+    }
+    if(![newSource isEqualToString:_source]){
+        _source=newSource;
+        if(_source){
+            [_webview loadHTMLString:_source baseURL:nil];
+        }
+    }
+    
 }
 
 - (void)loadURL:(NSString *)url
--- a/ios/sdk/WeexSDK/Sources/Component/WXWebComponent.m
+++ b/ios/sdk/WeexSDK/Sources/Component/WXWebComponent.m
@@ -51,6 +51,9 @@
 
 @property (nonatomic, strong) NSString *source;
 
+// save source during this initialization
+@property (nonatomic, strong) NSString *inInitsource;
+
 @property (nonatomic, assign) BOOL startLoadEvent;
 
 @property (nonatomic, assign) BOOL finishLoadEvent;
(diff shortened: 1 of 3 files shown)


ios/sdk/WeexSDK/Sources/Component/WXWebComponent.m
```diff
@@ -69,6 +71,11 @@ - (instancetype)initWithRef:(NSString *)ref type:(NSString *)type styles:(NSDict
 {
     if (self = [super initWithRef:ref type:type styles:styles attributes:attributes events:events weexInstance:weexInstance]) {
         self.url = attributes[@"src"];
+        
+        if(attributes[@"source"]){
+            self.source = attributes[@"source"];
```
acton393: the init method is executed in background thread named component method and  `self.source` will call [self setSource:""], 
the method for `setSource` called view operation which is only for main thread.

A better way is that define another member for source to save the variable during  this initialization,  and call  self.source in component hook lifeCycle like  `viewDidLoad`.

ios/sdk/WeexSDK/Sources/Module/WXWebViewModule.m
```diff
@@ -27,6 +27,7 @@ @implementation WXWebViewModule
 @synthesize weexInstance;
```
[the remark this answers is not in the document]
tw93: Thinks @acton393 , I have modified it

ios/sdk/WeexSDK/Sources/Component/WXWebComponent.m
```diff
@@ -104,6 +142,10 @@ - (void)updateAttributes:(NSDictionary *)attributes
     if (attributes[@"src"]) {
         self.url = attributes[@"src"];
     }
+
+    if (attributes[@"source"]) {
```
acton393: maybe we should update the initSource when update attributes,
as the view can be unloaded during the screen scroll, so the viewDidLoad method can called again with initSource

tw93: done!

PR #1051, opened 2018-03-07
[WEEX-241][iOS] add WXVideoComponent "poster" attribute

xuyouyang:
+ [iOS] add WXVideoComponent poster feature and add support "poster" attribute to <video>
testcase: http://dotwe.org/vue/4bf373d06d3b8021441a88b7f841ef3b

feature:241

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Component/WXVideoComponent.m
+++ b/ios/sdk/WeexSDK/Sources/Component/WXVideoComponent.m
@@ -89,13 +89,6 @@
         }
         
         [self addSubview:_playerViewController.view];
-        
-        _posterImageView = [[UIImageView alloc] init];
-        _posterImageView.userInteractionEnabled = YES;
-        [_posterImageView addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(posterTapHandler)]];
-        _posterImageView.hidden = YES;
-        [self addSubview:_posterImageView];
-        [self bringSubviewToFront:_posterImageView];
     }
     return self;
 }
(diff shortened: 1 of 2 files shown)


acton393:
hi,  @xuyouyang   what's the "poster"  feature，  is it  the paused picture about  player display?

xuyouyang:
NO.
The poster attribute specifies an image to be shown while the video is downloading, or until the user hits the play button. If this is not included, the first frame of the video will be used instead.
You can see this "https://www.w3schools.com/tags/att_video_poster.asp"
It is a attribute in HTML <video> tag

acton393:
@xuyouyang  I got it and find the w3c rules about it : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

thanks, we can make it more standard by add more attributes and styles, what about Android?

ios/sdk/WeexSDK/Sources/Component/WXVideoComponent.m
```diff
@@ -85,6 +89,13 @@ - (id)init
         }
         
         [self addSubview:_playerViewController.view];
+        
+        _posterImageView = [[UIImageView alloc] init];
```
acton393: How about lazy load when front-end developer specify the poster attributes

acton393:
@xuyouyang   as the [development process](http://weex.apache.org/development-process.html) shows, we need create a jira issue in apache, and format your pull request title and commit log.

   I have created a jira issue for this feature, and the issue id is WEEX-241, the issue address is https://issues.apache.org/jira/browse/WEEX-241

what you need to do next is to update the title as the following format 
 `[WEEX-jiraId][iOS/Android] summary`

thanks

acton393:
@xuyouyang  It look good to me and it will be accepted, thank you.
 It would be better if Android player can support poster attributes.

xuyouyang:
@acton393 Ha, it's sorry that I am not familiar with Android develop very much.😂

PR #1058, opened 2018-03-08
[WEEX-233] [android] More enhanced about <web> component 

TheKingOfGods:
Hi, It's solution to enhance <web> component in android, and the api was designed as same as IOS. More detail please look at  this [pr](https://github.com/apache/incubator-weex/pull/1047).

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/view/WXWebView.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/view/WXWebView.java
@@ -21,11 +21,19 @@ package com.taobao.weex.ui.view;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Color;
+import android.net.Uri;
 import android.net.http.SslError;
+import android.os.Build;
+import android.os.Handler;
+import android.os.Message;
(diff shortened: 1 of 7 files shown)


android/sdk/src/main/java/com/taobao/weex/ui/view/WXWebView.java
```diff
@@ -230,7 +295,79 @@ public void onReceivedTitle(WebView view, String title) {
                 }
             }
 
+            @Override
```
misakuo: Can we use `addJavascriptInterface` or `console.log` as the communication channel? `window.prompt` has some known memory performance issues in the case of high frequency call

TheKingOfGods: Yes, `addJavascriptInterface` is better, and it will be used when Android API >= 17 . But there is a [risk](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2012-6636) before 17. So `prompt`  is used as a downgrade when Android API < 17

misakuo:
This pull request has been merged, please close it manually 🤝 @TheKingOfGods

PR #1072, opened 2018-03-15
[WEEX-249][iOS]fix WXRuleManager multi-thread crash

xuyouyang:
fontFamilyDic type is WXThreadSafeMutableDictionary in WXUtility.m but the real type is NSMutableDictionary created in WXRuleManager.m.

It may cause crash in multi-thread.

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Manager/WXRuleManager.m
+++ b/ios/sdk/WeexSDK/Sources/Manager/WXRuleManager.m
@@ -89,7 +89,7 @@ static WXRuleManager *_sharedInstance = nil;
                 return;
             }
             if (!fontFamily) {
-                fontFamily = [WXThreadSafeMutableDictionary dictionary];
+                fontFamily = [[WXThreadSafeMutableDictionary alloc] init];
             }
             NSURL *fontURL = [NSURL URLWithString:fontSrc];
             if (!fontURL) {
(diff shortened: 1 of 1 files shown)


acton393:
hi @xuyouyang  thank you for your contribution, sure that's a bug, you are very careful. and we can change the variable type in WXUtility.m or change the type during its creation to fix it.

I found that `fontFamilyDic` is a second dictionary which is a object in `fontStorage` which is also `WXThreadSafeDictionary` and its key is 'fontFamily', I'm concerning about that if we change the `fontFamilyDic` type to `WXThreadSafeDictionary`, there can be a deadlock as `WXThreadSafeDictionary` using lock to ensure its thread-safe.


so I think we can fix it by changing type in `WXUtility.m`.

xuyouyang:
hi @acton393 if we change fontFamilyDic type to NSMutableDictionary，will it also cause multi-thread problem?
For example, other thread operate object in fontFamilyDic.

And I have a doubt that why WXThreadSafeDictionary can be a deadlock? It created in different queue and different lock.

acton393:
@xuyouyang  Cases occurs to me that one of thread operating the storage and then operating the fontFamilyDic, as they was in a same dictionary.  And for `os_unfair_lock`,  this is not a type of recursive lock, so it can not be acquired multi times.

PR #1076, opened 2018-03-21
[Weex-260][iOS]switch supports setting color

tw93:
## Background
- Currently, the [`<switch>`](http://weex-project.io/cn/references/components/switch.html) component in Weex only support the default green color, But many times the business needs to set color for the theme tone of the product, So <switch> needs to support setting the color.

## New Support &&  Demo
- tintColor: Background color when the switch is turned on.
- onTintColor: Border color and background color on Android when the switch is turned off.
- thumbTintColor: Color of the foreground switch grip.

<table><tr><td></td><td></td></tr></table>

## How to Use

```
<switch on-tint-color="#C33D3E" 
        thumb-tint-color="#FF7703" 
        tint-color="#850B0B" 
        checked="true"></switch>
```

#### Code
* Weex iOS:
  * [WXSwitchComponent.m](https://github.com/tw93/incubator-weex/blob/ios-feature-switch/ios/sdk/WeexSDK/Sources/Component/WXSwitchComponent.m)
* Weex Vue Demo:
  * [http://dotwe.org/vue/99a13cce5a429c7b9bce9bea24253935](http://dotwe.org/vue/99a13cce5a429c7b9bce9bea24253935)
  * [Demo Bundle JS](http://dotwe.org/raw/dist/99a13cce5a429c7b9bce9bea24253935.bundle.wx)

#### Details

It can be overwritten with the `onTintColor`、`thumbTintColor`、`tintColor` attributes of `switchView`, as well as the job of updating the attributes.

```
...
_switchView.onTintColor = _onTintColor;
_switchView.thumbTintColor = _thumbTintColor;
_switchView.tintColor = _tintColor;
```

Welcome to put forward any suggestion about the solution or other requirements for the <switch> component, Thanks！

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Component/WXSwitchComponent.m
+++ b/ios/sdk/WeexSDK/Sources/Component/WXSwitchComponent.m
@@ -36,6 +36,16 @@
 @property (nonatomic, assign)   BOOL    checked;
 @property (nonatomic, assign)   BOOL    disabled;
 
+//Background color when the switch is turned on.
+@property (nonatomic, strong)  UIColor *onTintColor;
+
+
+//Color of the foreground switch grip.
+@property (nonatomic, strong)  UIColor *thumbTintColor;
+
+//Border color and background color on Android when the switch is turned off
+@property (nonatomic, strong)  UIColor *tintColor;
+
 @end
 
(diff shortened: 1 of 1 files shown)


acton393:
it looks good to me , accepted

PR #1090, opened 2018-03-28
[WEEX-267][ios]monitor the frequency of module invoke

CYJB:
Add new optional method for module invoke monitor, do not use alarm anymore.

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Protocol/WXAppMonitorProtocol.h
+++ b/ios/sdk/WeexSDK/Sources/Protocol/WXAppMonitorProtocol.h
@@ -56,6 +56,6 @@
 
 @optional
 
-- (void)commitMonitorWithPage:(NSString *)pageName withPoint:(NSString *)monitorPoint withArgs:(NSDictionary *)args;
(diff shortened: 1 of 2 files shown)


ios/sdk/WeexSDK/Sources/Protocol/WXAppMonitorProtocol.h
```diff
@@ -54,4 +54,8 @@
 
 - (void)commitAppMonitorAlarm:(NSString *)pageName monitorPoint:(NSString *)monitorPoint success:(BOOL)success errorCode:(NSString *)errorCode errorMsg:(NSString *)errorMsg arg:(NSString *)arg;
 
+@optional
+
+- (void)commitMonitorWithPage:(NSString *)pageName withPoint:(NSString *)monitorPoint withArgs:(NSDictionary *)args;
```
acton393: maybe
` - (void)commitMonitorWithPage:(NSString *)pageName monitorPoint:(NSString *)monitorPoint args:(NSDictionary *)args   `
 is better?

CYJB: Yes, that is better

PR #1105, opened 2018-04-09
[WEEX-282][iOS] update layout system to support rtl direction

win80540:
Update WXComponent+layout.m to support CSS "direction:rtl".
Few languages such as Arabic, Hebrew, or Persian are written from Right to Left, but weex not support RTL layouts. To handle them. Since this PR merged, we can use weex for RTL languages.

Modified WXComponent+Layout.m:
	1.set direction value to  cssNode
	2.setNeedLayout changed to trigger sibling component recalculate cssNode->layout

Modified WXConvert.h/m:
	add convert for "direction"

Modified WXConvertTests.m:
	add Tests for convert direction


acton393:
@win80540   thanks for you contribution

PR #1110, opened 2018-04-12
 [WEEX-288][android] bug fix, fix on some case cannot get bundle type

yuhun-alibaba:
[WEEX-288][android] bug fix, fix on some case cannot get bundle type


Darin726:
LGTM

PR #1114, opened 2018-04-17
*[Android] box-shadow size should cover padding and border

Yanjiie:
Hi ~ I found some problem in Android side when i use box-shadow. Look that:

**In H5 side:**

**In iOS side:**

**In Android side:**

Now, you see the bug.
The box-shadow style should cover box's padding and border when render the shadow. But Android side is different from the iOS and H5 side.
These change fix this Bug.

This is a test case. [android-bugfix-boxshadow-testcase](http://dotwe.org/vue/4ab90f1cae7c5313c31ab4316b306a6e)


misakuo:
Nice job!

PR #1124, opened 2018-04-23
[jsfm] Support import script dynamically at runtime

Hanks10100:
The code here partially supports the feature of import script dynamically at runtime. It needs native render engines to implement the corresponding features to fully support it.

## API

```typescript
weex.importScript(url: string): Promise
```

Import javascript code from remote and run it in the current context. It takes a url string as the only argument and return a Promise, which will be fullfilled once the code is successfully executed.

## Usage

```js
weex.importScript('https://whatever.com/shared.js')
    .then(() => {
        console.log('Import and run script success!')
    })
    .catch(() => {
        console.log('Failed to import script dynamically')
    })
```


runtime/api/WeexInstance.js
```diff
@@ -128,6 +129,31 @@ export default class WeexInstance {
     return null
   }
 
+  importScript (src, options = {}) {
+    const id = getId(this)
+    const taskCenter = getTaskCenter(id)
+    return new Promise((resolve, reject) => {
+      if (!taskCenter || typeof !taskCenter.send === 'function') {
```
imyzf: `typeof !taskCenter.send === 'function' ` should be `typeof taskCenter.send !== 'function'`

PR #1127, opened 2018-04-23
[WEEX-301][iOS] WXImageComponent: set imageview in completion block

xiayun200825:
#### bugfix
1. Apply image to imageview in completion block while downloading via url.


xuyouyang:
@xiayun200825 thank you for your contributions. 
Can you give us some testcase through 'http://dotwe.org/vue' ?

PR #1138, opened 2018-04-27
[WEEX-299][android] Touch event will transmit to next layer, requestDisallowInterceptTouchEvent should be reset for every touch event

gubaojian:
[WEEX-299][android] Touch event will transmit to next layer, requestDisallowInterceptTouchEvent should be reset for every touch event


misakuo:
nice

PR #1146, opened 2018-05-04
[iOS]bugfix: toast should display on weex window

x461909066:
bugfix: weex toast should display on the window that contain weex root view


x461909066:
bugfix: weex toast should display on the window that contain weex root view

acton393:
hi @x461909066   please follow the development-process  here, pay more attention to the commit message format http://weex.apache.org/development-process.html

PR #1162, opened 2018-05-10
[WEEX-349][jsengine] Make weex core can run on single process mode

Darin726:
[WEEX-349][jsengine] Make weex core can run on single process mode

Since Weex core may fork multi process failed, so we make it can run
in single process when failed to run multi process.

Bug: WEEX-349

https://issues.apache.org/jira/browse/WEEX-349


yuhun-alibaba:
LGTM

PR #1175, opened 2018-05-15
[WEEX-363][Android] Fix callNativeModule arguments NPE.

Kelvin-May:


gubaojian:
please rebase new code, then do the update

PR #1195, opened 2018-05-22
* [Core] Reorganized the CMake build files.

jondong:
This patch fixed some of the issues in CMake build config and related
files, which includes:

1. Renamed the project as "WeexCore", also renamed the corresponding
  libraries as libweexcore and libweexipc, which better described
  the project current status.
2. Splitted the source files into COMMON_SRCS which are platform
  irrelevant, as well as ANDROID_SRCS which are platform relevant.
3. Removed some unnecessary includes and links.


jondong:
@zshshr @YorkShen PTAL, thanks

PR #1202, opened 2018-05-23
* [android] Removed unit test task from weex_sdk travis build.

jondong:
As unit test codes for Weex sdk android stays unmaintained for long,
which breaks unit tests constantly, and need to rewrite. So we decided
to change the behavior of the travis build, we will only verify
Weex sdk build instead of running the unit test cases, until the unit
test codes get rewrite. Also changed the corresponding text output.

I'm sure this patch will make travis happy.

CheckList:

**if** *isBugFix* **:**

  * [ ] Make sure that you add at least one unit test for the bug which you had fixed.

**elif** *isNewFeature* **:**


jondong:
@zshshr @YorkShen PTAL.

jondong:
rebased, please help to check and merge if applicable. thanks. @YorkShen @zshshr

PR #1221, opened 2018-05-29
[WEEX-435][android] modify playground app camera runtime permission request behavior

shawnlinboy:
For `ActivityCompat.shouldShowRequestPermissionRationale()`, it means your app should display a dialog or something else to explain why you are going to request this permission, and **try again to request the permission**, see https://developer.android.com/training/permissions/requesting and https://github.com/googlesamples/android-RuntimePermissionsBasic/blob/master/Application/src/main/java/com/example/android/basicpermissions/MainActivity.java#L124

For the original code, if the permission was **NOT** granted, the `ContextCompat.checkSelfPermission(mWXSDKInstance.getContext(), Manifest.permission.CAMERA) != PackageManager.PERMISSION_GRANTED` will always returns `true` and `ActivityCompat.shouldShowRequestPermissionRationale((Activity) mWXSDKInstance.getContext(), Manifest.permission.CAMERA)` did the same thing, which means If a user tap `扫描二维码`, the `Weex playground need the camera permission to scan QR code` toast will always show up without prompting a permission request dialog, users have to navigate to system app management and grant permission manually. If one user was not familiar with this mechanism,the Playground app will always be not able to scan code.

This patch fix can show a dialog to tell user that `Weex playground need the camera permission to scan QR code` **and then request this code immediately**, it will make the a system permission dialog alert and the user can now grant permission directly.

### DIFF
--- a/android/playground/app/src/main/java/com/alibaba/weex/IndexActivity.java
+++ b/android/playground/app/src/main/java/com/alibaba/weex/IndexActivity.java
@@ -149,11 +149,14 @@ public class IndexActivity extends AbstractWeexActivity {
   @Override
   public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
     super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-    if (requestCode == CAMERA_PERMISSION_REQUEST_CODE && grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
-      startActivity(new Intent(this, CaptureActivity.class));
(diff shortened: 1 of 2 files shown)


xuyouyang:
@shawnlinboy Thank you for your contribution！！As the [development process shows](http://weex.apache.org/development-process.html), we need create a jira issue in [apache](https://issues.apache.org/jira/browse/WEEX), and format your pull request title and commit log. :)

shawnlinboy:
@xuyouyang Thanks for reviewing, already created one.
 https://issues.apache.org/jira/browse/WEEX-435

gubaojian:
in your onRequestPermissionsResult you don't handle WRITE_EXTERNAL_STORAGE_PERMISSION_REQUEST_CODE and give an tips. please add this case.

shawnlinboy:
@gubaojian They didn't handle that so I'd like not to add extra code to that.

PR #1253, opened 2018-06-08
[WEEX-419][android] weeks bugfix for security check

gubaojian:
[WEEX-419][android] weeks bugfix for security check


YorkShen:
Seems like already merged.

PR #1254, opened 2018-06-08
[WEEX-445][jsfm] export requireModule to global

imyzf:

CheckList:

**if** *isBugFix* **:**

  * [ ] Make sure that you add at least one unit test for the bug which you had fixed.

**elif** *isNewFeature* **:**

This is a API for internal usage, so no test and no doc update.

### DIFF
--- a/runtime/api/init.js
+++ b/runtime/api/init.js
@@ -117,6 +117,7 @@ function createInstanceContext (id, options = {}, data) {
   Object.assign(runtimeContext, services, {
     weex,
     getJSFMVersion,
+    requireModule: (...args) => weex.requireModule(...args),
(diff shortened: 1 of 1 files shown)


runtime/api/init.js
```diff
@@ -117,6 +117,7 @@ function createInstanceContext (id, options = {}, data) {
   Object.assign(runtimeContext, services, {
     weex,
     getJSFMVersion,
+    requireModule: weex.requireModule,
```
Hanks10100: I think using `(..args) => weex.requireModule(...args)` is more reliable. We should not assume that the context of `requireModule` has bind to `weex` even if it actually does.

imyzf: done.

PR #1291, opened 2018-06-26
[WEEX-479] Update jsbundle of weex playground app

erha19:
Fixed jira issue: https://issues.apache.org/jira/browse/WEEX-479


erha19:
@Hanks10100 needs you review

Hanks10100:
The new files are not compressed. Every thing else looks fine for me.

PR #1292, opened 2018-06-26
[android][WEEX-474] when the app's WXInstance destroy, but Weex page call toast Toast ,crash

gubaojian:
[android][WEEX-474] when the app's WXInstance destroy, but Weex page call toast Toast ,crash


YorkShen:
Seems like already merged

PR #1295, opened 2018-06-26
[Relese] Fixed the sed issue

jondong:
Use Mac sed to generate release source packages instead of using
gnu-sed, if you installed gnu-sed on your Mac.


jondong:
@cxfeng1

PR #1299, opened 2018-06-28
[WEEX-483][Android] get options from arguments in callNativeModule

Darin726:
https://issues.apache.org/jira/browse/WEEX-483
[WEEX-483][Android] get options from arguments in callNativeModule

options has been put in the arguments, So we should get options from arguments

Bug: WEEX-483


gubaojian:
you should handle the arguments in your javascript code, not in weex sdk

PR #1301, opened 2018-06-29
[WEEX-485][Android] fix slider bug: scroll when scrollable=false make…

luckybilly:
[WEEX-485][Android] fix slider bug: scroll when scrollable=false make index not work

issue: [[WEEX-485]](https://issues.apache.org/jira/browse/WEEX-485)


luckybilly:
PR has been merged, but the status is not updated... Is it a github bug?

PR #1310, opened 2018-07-04
[WEEX-490][android] fix the bug of input component when setting prope…

yangchendev:
…rty "type=tel" and "type=number" in input component，users can also input text

Bug: 490


WaldenCael:
two input in android, one type is text, other is password, when switch cursor in tow input, android keyboard will be disappear, keyboard will appear after touch input two times . can you help to fix this? thanks!

PR #1330, opened 2018-07-16
[android]update zxing core to gradle dependency

gengjiawen:
Replace zxing core with gradle dependency


gengjiawen:
@YorkShen Can you review this pr, thanks.

PR #1353, opened 2018-07-20
[WEEX-525] Created the automate beta branch-off scripts.

jondong:
The script beta-branchoff.sh is to automate the weekly beta release
branch-off procedure. It helps to identify which is the latest
non-RC version number and release a new beta branch with the number.
The branch will be branched-off from origin/master and will be pushed
onto apache after its creation.

A valid beta version number looks like: 0.18.0-beta1
A Valid beta branch name looks like: 0.18.0-beta1-180720

common.sh provides some useful gadgets for the scripts and can be
shared across scripts.

### DIFF
--- a/scripts/beta-branchoff.sh
+++ b/scripts/beta-branchoff.sh
@@ -1,0 +1,139 @@
+#!/usr/bin/env bash
+
+dir_name=$(dirname $0)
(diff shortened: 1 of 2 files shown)


jondong:
@Hanks10100 @cxfeng1 PTAL, thanks

PR #1511, opened 2018-09-12
[WEEX-611][Android] Fix java.lang.ArrayIndexOutOfBoundsException

Darin726:
[WEEX-611][Android] Fix java.lang.ArrayIndexOutOfBoundsException
arrayList is not thread safe, if arraylist add item in multithread may produce ArrayIndexOutOfBoundsException
Bug: WEEX-611


android/sdk/src/main/java/com/taobao/weex/utils/batch/BatchOperationHelper.java
```diff
@@ -31,7 +32,7 @@
 
 
   private BactchExecutor mExecutor;
-  private ArrayList<Runnable> sRegisterTasks = new ArrayList<>();
+  private CopyOnWriteArrayList<Runnable> sRegisterTasks = new CopyOnWriteArrayList<>();
```
YorkShen: `CopyOnWriteArrayList` is bad for performance in high frequency invoke situation. Be careful

PR #1520, opened 2018-09-13
[jsfm] separate vue framework and add vue-api

imyzf:
Separate vue framework from js framework, and weex-vue-api.js should be executed before every bundle.

### DIFF
--- a/runtime/entries/vue.js
+++ b/runtime/entries/vue.js
@@ -28,10 +28,7 @@ const Vue = {
     }
 
     const instanceContext = {
-      __weex_instance_id__: instanceId,
-      __weex_document__: weex.document,
       __weex_data__: data || {},
-      __weex_config__: weex.config,
       __instance_lifecycle__: instance.lifecycle
     };
 
--- a/runtime/entries/vue.js
+++ b/runtime/entries/vue.js
@@ -17,7 +17,45 @@
  * under the License.
  */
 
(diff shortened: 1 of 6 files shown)


runtime/entries/vue.js
```diff
@@ -17,7 +17,45 @@
...
+const Vue = {
+  createInstanceContext (instanceId, runtimeContext, data) {
+    const weex = runtimeContext.weex
+    const instance = instanceOptions[instanceId] = {
+      document: weex.document,
+      lifecycle: {}
+    }
+
+    const instanceContext = {
+      __weex_instance_id__: instanceId,
```
Hanks10100: The `__weex_instance_id__`, `__weex_document__`, `__weex_config__` can be get from the `weex` variable, there is no need to pass those apis to instance context.

runtime/entries/vue.js
```diff
@@ -17,7 +17,45 @@
  * under the License.
  */
 
-import setup from './setup'
-import * as Vue from 'weex-vue-framework'
+const instanceOptions = {}
```
Hanks10100: The `instanceOptions` variable will always exist even pages are destroyed, it may cause a potential memory leak, we should figure out another way to implement instance lifecycles.

Hanks10100:
This PR is closed because it is outdated.

PR #1522, opened 2018-09-13
[WEEX-612][android] Fixed after input component default set  a string…

tody123:
[WEEX-612] Fixed after input component default set a string on android platform, and then remove it, not trigger input event.
see more here https://issues.apache.org/jira/browse/WEEX-612

Bug:612


zshshr:
@yxping  review and  check the PullRequest

yxping:
LGTM

viveksinha9:
Status of WEEX-612 is still open. When this fix will be released?

PR #1550, opened 2018-09-20
Performance info is not an error, change it to debug info

or0fun:
Performance info is not an error, change it to debug info.


YorkShen:
When you actual release your app, the debug level log is stripped, but the *Performance* is important and we'd like to print it even if it is a release app.

I will close PR. Thansk

PR #1645, opened 2018-10-18
[core]fix string operating number

holyace:
fix OP_DIV operation when one of the operation of the operator is a string.


YorkShen:
This PR will be closed later as vm.cc will be moved to another repose

PR #1644, opened 2018-10-17
[WEEX-648][iOS]native batch for dom operations generating from JavaScript

acton393:
introduce a native batch for dom operations generating from JavaScript
  wrap a series of dom operations with beginBatch and endBatch directives, when every V-sync signal comes, we try to ensure that the operations between beginBatch and endBatch can be performed in current V-sync, this policy can drop frames maybe, for our policy dropping frames, we only allow 20 frames at most.

see Android policy pull request #1647 

feature:648

PR #1647, opened 2018-10-18
[WEEX-648][Android]native batch for dom operations generating from JavaScript

acton393:
introduce a native batch for dom operations generating from JavaScript
wrap a series of dom operations with beginBatch and endBatch directives, when every V-sync signal comes, we try to ensure that the operations between beginBatch and endBatch can be performed in current V-sync, this policy can drop frames maybe, for our policy dropping frames, we only allow 20 frames at most.
see iOS pull request https://github.com/apache/incubator-weex/pull/1644

feature:648

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/module/WXDomModule.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/module/WXDomModule.java
@@ -27,6 +27,8 @@ import com.taobao.weex.dom.binding.JSONUtils;
 import com.taobao.weex.ui.action.ActionAddRule;
 import com.taobao.weex.ui.action.ActionGetComponentRect;
 import com.taobao.weex.ui.action.ActionInvokeMethod;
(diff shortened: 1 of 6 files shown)


android/sdk/src/main/java/com/taobao/weex/ui/module/WXDomModule.java
```diff
@@ -118,6 +123,16 @@ public Object callDomMethod(String method, JSONArray args, long... parseNanos) {
           }
           new UpdateComponentDataAction(mWXSDKInstance, args.getString(0), JSONUtils.toJSON(args.get(1)), args.getString(2)).executeAction();
           break;
+        case BATCH_BEGIN: {
+          String ref = args.size() >= 1 ? args.getString(0) : null;
```
Darin726: args may equals null

acton393: @Darin726   thanks, I have checked value for args, just return if null

PR #1662, opened 2018-10-23
[iOS] treat 750 as container width not device width when rotated or splited in ipad

lynnleelhl:
If a component's width is set to 750, weex will calculate it's width as the device width, however when you put weex in ipad, the size of weex container don't usually equal to device width ( e.g. split view ), and also the device width didn't update after rotation. This PR solved the problem.


lynnleelhl:
I found out some problem.

PR #1695, opened 2018-11-02
Fixed 'index' is invalid when 'slider' data reduction

ttarigo:


wqyfavor:
Thanks, I think your pull request is OK and I will merge it. But could you also provide a demo for this case?

PR #1733, opened 2018-11-09
Update WXBridgeManager.m

HJaycee:
options nullable is better for developers


wqyfavor:
We accept your advice. But I will close your pr and create another one next week.

PR #1741, opened 2018-11-12
[devtool] Updated the Apache release scripts.

jondong:
Added 'npm run build:rax' to generate Rax precompiled js files, and
updated the build command and generated project for iOS.

### DIFF
--- a/scripts/build_from_source.sh
+++ b/scripts/build_from_source.sh
@@ -18,20 +18,22 @@
 
 set -e
 if [ ! -f scripts/build_from_source.sh ];then
-    echo "This script must execute in project root"
+    echo "This script must be executed from project root."
     exit 1
 fi
 
-echo "Build Weex SDK From Source...."
+echo "Build Weex SDK from source...."
 
 npm install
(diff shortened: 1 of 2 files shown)


jondong:
@YorkShen @Hanks10100 @cxfeng1 PTAL. thanks.

scripts/build_from_source.sh
```diff
@@ -18,20 +18,22 @@
...
+npm run build:rax
 
 echo "Weex JS Framework build completed."
 sleep 2
 
-mkdir pre-build
+# Copy built JS resources to their destination.
+[ -d pre-build ] || mkdir pre-build
 cp dist/weex-js-framework.min.js pre-build/native-bundle-main.js
 cp dist/weex-js-framework.min.js android_sdk/assets/native-bundle-main.js
```
Hanks10100: I think the file structure had been refactored, here should be `android/sdk/assets` instead of `android_sdk/assets`.

Hanks10100: Got it! I mixed up those folders.

PR #1800, opened 2018-11-22
[IOS]add call onJSException when error log(#1767)

zxchehe:


wqyfavor:
Sorry, we couldn't call commitCriticalExceptionRT in this method. Besides, logs with level of WXLogFlagWarning and WXLogFlagError are handled by appMonitorHandler, you could change your implementation of WXAppMonitorProtocol.

PR #1847, opened 2018-11-29
[Android] fix playground app crash while url query is null

erha19:
Changed:

- fix playground app crash while url query is null.

Fixed https://github.com/apache/incubator-weex/issues/1842.


android/playground/app/src/main/java/com/alibaba/weex/WXPageActivity.java
```diff
@@ -275,7 +275,7 @@ private void startHotRefresh() {
       URL url = new URL(mUri.toString());
       String host = url.getHost();
       String query = url.getQuery();
-      String wsport = getUrlParam("wsport", query);
+      String wsport = query == null? "8082" : getUrlParam("wsport", query);
```
YorkShen: TextUtils.isEmpty(query)? "8082" : getUrlParam("wsport", query)

erha19:
@YorkShen done!

PR #1909, opened 2018-12-14
[jsfm] attach data to WeexInstance

imyzf:


runtime/api/WeexInstance.js
```diff
@@ -53,9 +53,10 @@ function moduleSetter (id, module, method, fn) {
 }
 
 export default class WeexInstance {
-  constructor (id, config) {
+  constructor (id, config, data) {
     setId(this, String(id))
     this.config = config || {}
+    this.data = data || {}
```
Hanks10100: Using `data` as the property is too common and ambiguous. Maybe you should use a more explicit and semantic name here.

PR #1964, opened 2018-12-25
[weex][android] fix the custom file ttf not work well, need replace u…

brucetoo:


android/sdk/src/main/java/com/taobao/weex/utils/FontDO.java
```diff
@@ -97,7 +97,12 @@ private void parseSrc(String src, WXSDKInstance instance) {
           mType = TYPE_NETWORK;
         } else if (Constants.Scheme.FILE.equals(scheme)) {
           mType = TYPE_FILE;
-          mUrl = uri.getPath();
+            /**
+             * eg: file://name/A/B.ttf
```
YorkShen: I'm not quite understand your case. What's problem with `/name/A/B.ttf`

brucetoo: like what [official doc](http://weex-project.io/cn/references/modules/custom_font.html) said :  with `url('file://storage/emulated/0/A/B/**.ttf` to custom font by file path, but the result of path format that will like `emulated/0/A/B/**.ttf` ,rather than `storage/emulated/0/A/B/**.ttf` which will cause `File Not Found Exception`. Because `getPath` format schema `storage` as **authority** , as you can see `[scheme:][//authority][path][?query][#fragment]`

YorkShen: In this case, I think `getSchemeSpecificPart` is enough

YorkShen: In that case, I think `getSchemeSpecificPart` would be better

brucetoo: @YorkShen I don't think so!
`getEncodedSchemeSpecificPart ` will keep the origin information whatever it looks like before
but `getSchemeSpecificPart` will decode the  raw information but real scenes where we need keep it original

PR #1978, opened 2018-12-28
Weex prerender

lvfen:
### Feature: Weex Prerender 
to enable create instance without create real view


gubaojian:
can  you provide your ding talk or we chat talk, i want to know more detail about it, me we chat 18072871207

PR #2039, opened 2019-01-11
make android support anti-aliased image

hjmmc:
Fixed image component anti-alias problem when rendering some non-standard cut image under Android (ios have no such problem)

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java
+++ b/android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java
@@ -132,9 +132,10 @@ public class ImageDrawable extends PaintDrawable {
     if (Build.VERSION.SDK_INT == Build.VERSION_CODES.LOLLIPOP) {
       // fix api 21 PaintDrawable crash
       paint.setAntiAlias(false);
+    } else {
+      // make android support anti-aliased image
+      paint.setFlags(Paint.ANTI_ALIAS_FLAG|Paint.FILTER_BITMAP_FLAG);
     }
-    // add support android anti-aliasing image
-    paint.setFlags(Paint.ANTI_ALIAS_FLAG|Paint.FILTER_BITMAP_FLAG);
     super.onDraw(shape, canvas, paint);
   }
 
--- a/android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java
+++ b/android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java
@@ -49,6 +49,8 @@ public class ImageDrawable extends PaintDrawable {
               (bm = ((BitmapDrawable) original).getBitmap()) != null) {
         ImageDrawable imageDrawable;
         imageDrawable = new ImageDrawable();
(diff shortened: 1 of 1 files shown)


android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java
```diff
@@ -132,6 +132,9 @@ protected void onDraw(Shape shape, Canvas canvas, Paint paint) {
     if (Build.VERSION.SDK_INT == Build.VERSION_CODES.LOLLIPOP) {
       // fix api 21 PaintDrawable crash
       paint.setAntiAlias(false);
+    } else {
+      // make android support anti-aliased image
+      paint.setFlags(Paint.ANTI_ALIAS_FLAG|Paint.FILTER_BITMAP_FLAG);
```
YorkShen: paint.setAntiAlias(ture) will be enough, I think

hjmmc: ```paint.setAntiAlias(ture)``` is not enough.
I tested it on Android 8.0. Only with this line of code, the image doesn't appear jagged.

YorkShen:
Could you provide a demo on [dot.we](http://dotwe.org/vue) and a screenshot of your phone, as I don't find any anti-alias problem on Android

hjmmc:
[demo](http://dotwe.org/vue/9ccfc93e6ce1e0698917530487ff7f75)

screenshot in Xiaomi8(Android 9.0)

screenshot_2019-01-16-17-56-01-831_com alibaba we

YorkShen:
I reproduce the problem on my Android 9 phone, but it works well on my Android 6 phone, it seems like to a bug in Android system.

Anyway, I suggest you add the following code snippet in the [following line]:(https://github.com/YorkShen/incubator-weex/blob/2f8caedb147b13182647a1dab473ba9cb31e54e4/android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java#L51)

```       
 imageDrawable.getPaint().setFilterBitmap(true);
```

It works as expected on my Android 9 phone.

android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java
YorkShen: Please add `imageDrawable.getPaint().setFilterBitmap(true);` when creating `ImageDrawable`, no need to invoke setFlags during `draw()`

hjmmc:
> I reproduce the problem on my Android 9 phone, but it works well on my Android 6 phone, it seems like to a bug in Android system.
> 
> Anyway, I suggest you add the following code snippet in the [corresponding line](https://github.com/YorkShen/incubator-weex/blob/2f8caedb147b13182647a1dab473ba9cb31e54e4/android/sdk/src/main/java/com/taobao/weex/utils/ImageDrawable.java#L51):
> 
> ```
>  imageDrawable.getPaint().setFilterBitmap(true);
> ```
> It works as expected on my Android 9 phone.

okay

PR #2198, opened 2019-03-11
[iOS][Android] fix scrollToElement performance incorrect  offset

win80540:
Since i start use Rax UI Framework - Nuke, i found i can't scroll to correct element on RTL layout.
Because Nuke will add a  container div between scroller and scroller's child element. 

ScrollToElement only can scroll to the direct child element on RTL direction in the online version，
now i fix the problem，scrollToElement can scroll to indirect child element.

BTW. i found scroll to begin let front-end engineer confused, now i found a better way to key offset when layout direction changed.

Nuke DEMO： https://jsplayground.taobao.org/raxplayground/5bf4b469-8eb1-4f5d-91df-912cd085ab7b
Direction chang DEMO：http://dotwe.org/vue/09cb6e35412f96c25c423fd50a929597

scrollToElement

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
@@ -86,6 +86,7 @@ public class WXScroller extends WXVContainer<ViewGroup> implements WXScrollViewL
   private int mOffsetAccuracy = 10;
   private Point mLastReport = new Point(-1, -1);
   private boolean mHasAddScrollEvent = false;
+  private boolean mIslastDirectionRTL = false;
 
   private static final int SWIPE_MIN_DISTANCE = 5;
   private static final int SWIPE_THRESHOLD_VELOCITY = 300;
@@ -101,6 +102,8 @@ public class WXScroller extends WXVContainer<ViewGroup> implements WXScrollViewL
   private boolean pageEnable = false;
   private boolean mIsHostAttachedToWindow = false;
   private View.OnAttachStateChangeListener mOnAttachStateChangeListener;
+
+  private boolean mlastDirectionRTL = false;
 
   public static class Creator implements ComponentCreator {
     @Override
@@ -485,15 +488,13 @@ public class WXScroller extends WXVContainer<ViewGroup> implements WXScrollViewL
         final WXScroller component = this;
         final View.OnLayoutChangeListener listener = new View.OnLayoutChangeListener() {
           @Override
-          public void onLayoutChange(View view, int left, int top, int right, int bottom, int oldLeft, int oldTop, int oldRight, int oldBottom) {
+          public void onLayoutChange(View view, final int left, int top, final int right, int bottom, final int oldLeft, int oldTop, final int oldRight, int oldBottom) {
             final View frameLayout = view;
             scrollView.post(new Runnable() {
               @Override
               public void run() {
-                if (isLayoutRTL()) {
-                  int mw = frameLayout.getMeasuredWidth();
-                  scrollView.scrollTo(mw, component.getScrollY());
-                } else {
+                if (isLayoutRTL() != mIslastDirectionRTL) {
+                  mIslastDirectionRTL = isLayoutRTL();
                   boolean scrollToBegin = true;
                   Object scrollToBeginOnLTR = getAttrs().get("scrollToBegin");
                   if (scrollToBeginOnLTR instanceof String){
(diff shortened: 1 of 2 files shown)


android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
```diff
@@ -482,44 +485,45 @@ public void onScrollChanged(WXHorizontalScrollView scrollView, int x, int y, int
...
-                  }
+      final WXScroller component = this;
+      final View.OnLayoutChangeListener listener = new View.OnLayoutChangeListener() {
+        @Override
+        public void onLayoutChange(View view, final int left, int top, final int right, int bottom, final int oldLeft, int oldTop, final int oldRight, int oldBottom) {
+          final View frameLayout = view;
+          scrollView.post(new Runnable() {
+            @Override
+            public void run() {
+              if (mIslastDirectionRTL != null && isLayoutRTL() != mIslastDirectionRTL.booleanValue()) {
```
YorkShen: As mIslastDirectionRTL is `Boolean`, which could cause null pointer exception here,

Why don't use `boolean mIslastDirectionRTL` directly?

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
```diff
@@ -482,44 +485,45 @@ public void onScrollChanged(WXHorizontalScrollView scrollView, int x, int y, int
...
+                int totalWidth = getInnerView().getChildAt(0).getWidth();
+                int displayWidth = getInnerView().getMeasuredWidth();
+                scrollView.scrollTo(totalWidth - currentX - displayWidth, component.getScrollY());
+              } else if (isLayoutRTL()) {
+                // if layout direction not changed, but width changede, we need keep RTL offset
+                int oldWidth = oldRight - oldLeft;
+                int width = right - left;
+                int changedWidth = width - oldWidth;
+                if (changedWidth != 0) {
+                  scrollView.scrollBy(changedWidth, component.getScrollY());
```
YorkShen: You are missing an `else` case, which could be a problem here.

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
```diff
@@ -482,44 +485,45 @@ public void onScrollChanged(WXHorizontalScrollView scrollView, int x, int y, int
...
-            final View frameLayout = view;
-            scrollView.post(new Runnable() {
-              @Override
-              public void run() {
-                if (isLayoutRTL()) {
-                  int mw = frameLayout.getMeasuredWidth();
-                  scrollView.scrollTo(mw, component.getScrollY());
-                } else {
-                  boolean scrollToBegin = true;
-                  Object scrollToBeginOnLTR = getAttrs().get("scrollToBegin");
```
YorkShen: `ScrollBegin` is added in https://github.com/apache/incubator-weex/pull/2165

Is there an reason to remove?

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
```diff
@@ -798,19 +802,24 @@ public void scrollTo(WXComponent component, Map<String, Object> options) {
     int viewXInScroller = 0;
     if (this.isLayoutRTL()) {
       // if layout direction is rtl, we need calculate rtl scroll x;
-      if (getInnerView().getChildCount() > 0) {
-        int totalWidth = getInnerView().getChildAt(0).getWidth();
-        int displayWidth = getInnerView().getMeasuredWidth();
-        viewXInScroller = totalWidth - (component.getAbsoluteX() - getAbsoluteX()) - displayWidth;
+      if (component.getParent() != null && component.getParent() == this) {
```
YorkShen: Any reason to add this `if`?

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
```diff
@@ -798,19 +802,24 @@ public void scrollTo(WXComponent component, Map<String, Object> options) {
...
+          int totalWidth = getInnerView().getChildAt(0).getWidth();
+          int displayWidth = getInnerView().getMeasuredWidth();
+          viewXInScroller = totalWidth - (component.getAbsoluteX() - getAbsoluteX()) - displayWidth;
+        } else {
+          viewXInScroller = component.getAbsoluteX() - getAbsoluteX();
+        }
       } else {
-        viewXInScroller = component.getAbsoluteX() - getAbsoluteX();
+        int displayWidth = getInnerView().getMeasuredWidth();
+        viewXInScroller = component.getAbsoluteX() - getAbsoluteX() - displayWidth + (int)component.getLayoutWidth();
```
YorkShen: ` - displayWidth + (int)component.getLayoutWidth()`

What's purpose

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
win80540: if use boolean, default is false, is not what i want.
I need record direction when last layout change occur, it not always false, and i need distinct is layout change trigger first or not.

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
win80540: Because it do not scroll to begin anymore.

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
win80540: If changedWidth == 0,  we should do nothing

android/sdk/src/main/java/com/taobao/weex/ui/component/WXScroller.java
win80540: If hit this ```if```, it mean's scroll to indirect child element, then the offsetX calculate will different from direct child element

PR #2215, opened 2019-03-14
[iOS] Fix the border of view hasn't gone after changing border width to 0 dynamically; (#2153)

NickeyLin:
The point is `_needsDrawBorder` in [WXComponent needsDrawRect]. 

When the border width changes to 0 from a non-zero value, it calls `... --> display --> _willDisplayLayer: --> needsDrawRect --> ...`. Then the border drawing is according to the result of  `needsDrawRect`. 

```
- (void)_willDisplayLayer:(CALayer *)layer
{
    ...
    BOOL needsDrawRect = [self needsDrawRect];
    WXDisplayBlock displayBlock;
    if (_useCompositing) {
        displayBlock = [self _compositeDisplayBlock];
    } else {
        displayBlock = [self _displayBlock];
    }
    WXDisplayCompletionBlock completionBlock = [self _displayCompletionBlock];
    
    if (!displayBlock || !needsDrawRect) {
        if (completionBlock) {
            completionBlock(layer, NO);
        }
        return;
    }
    ...
}
```
And the result of `needsDrawRect` is related on the result of `_needsDrawBorder`.

```objc
- (BOOL)needsDrawRect
{
    if (_useCompositing || _isCompositingChild) {
        return YES;
    }
    
    if (![self _needsDrawBorder] ) {
        ...
        return NO;
    }
    
    return YES;
}

```

But the result of `_needsDrawBorder` is only calculated by comparing the current border variables(`_border...Style`, `_border...Color`, `_border...Width`) with the pre-defaults. So that when the border width changes to zero from a non-zero value, `_needsDrawBorder` returns `NO` because current vars are equal to the pre-defaults. Then that cause `_drawBorderWithContext` not be called. Border is still there.

I considered that `_needsDrawBorder` has it's practical use, so I did not change it's implementation, just add a variable for recording the last borders instead. 

Issue detail: #2153

### DIFF
--- a/ios/sdk/WeexSDK/Sources/Display/WXComponent+Display.m
+++ b/ios/sdk/WeexSDK/Sources/Display/WXComponent+Display.m
@@ -31,6 +31,15 @@
 
 #pragma clang diagnostic ignored "-Wobjc-protocol-method-implementation"
 
+typedef NS_ENUM(NSInteger, WXComponentBorderRecord) {
+    WXComponentBorderRecordNone = 0,
+    WXComponentBorderRecordTop = 1,
+    WXComponentBorderRecordRight = 1 << 1,
+    WXComponentBorderRecordBottom = 1 << 2,
+    WXComponentBorderRecordLeft = 1 << 3,
+    WXComponentBorderRecordAll = WXComponentBorderRecordTop | WXComponentBorderRecordRight | WXComponentBorderRecordBottom | WXComponentBorderRecordLeft
+};
+
 @implementation WXComponent (Display)
 
 #pragma mark Public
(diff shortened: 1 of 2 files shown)


NickeyLin:
@wqyfavor  Please check the code and CocoaPods can be used after passing.  tks.

ios/sdk/WeexSDK/Sources/Display/WXComponent+Display.m
```diff
@@ -61,7 +70,7 @@ - (BOOL)needsDrawRect
         return YES;
     }
     
-    if (![self _needsDrawBorder]) {
+    if (![self _needsDrawBorder] && _lastBorderRecords == WXComponentBorderRecordNone) {
```
NickeyLin: Return NO only if there is no border for the last time and this time.

Issue #2253, opened 2019-03-27
[Android] Get blank page when JS exception raised before batch end.


sunshl:
We get blank page when JS exception raised before batch end.
In android/sdk/src/main/java/com/taobao/weex/ui/WXRenderManager.java, the batch update operations should be saved in different containers between different pages.

YorkShen:
The problem seems solved, thanks for your contribute.

PR #2254, opened 2019-03-27
[Android] Fix blank page when JS exception raised before batch end.

sunshl:
* bugfix: Fix blank page when JS exception raised before batch end.
* issue: https://github.com/apache/incubator-weex/issues/2253

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/WXRenderManager.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/WXRenderManager.java
@@ -165,6 +165,8 @@ public class WXRenderManager {
       return;
     }
 
+    // If more than two pages exist and mCurrentBatchInstanceId not matches new instanceId, we will post all stashed actions at once.
+    // That will cause losing efficacy of batch action, but it is acceptable because it's not serious problem.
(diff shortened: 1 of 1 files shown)


android/sdk/src/main/java/com/taobao/weex/ui/WXRenderManager.java
```diff
@@ -144,6 +165,15 @@ public void postGraphicAction(final String instanceId, final BasicGraphicAction
       return;
     }
 
+    if (mCurrentBatchInstanceId != null && instanceId != null && !mCurrentBatchInstanceId.equals(instanceId) && mBatchActions.size() > 0) {
```
YorkShen: Will your change makes `public void postGraphicAction(final String instanceId, final BasicGraphicAction action)` behaviour unexpectedly if there are two weex instance as you only use one field called `mCurrentBatchInstanceId`?

sunshl: If more than two pages exist and mCurrentBatchInstanceId not matches new instanceId, we will post all stashed actions at once.

That will cause losing efficacy of batch action, but it is acceptable because it's not serious problem.

PR #2279, opened 2019-04-04
[Android] fix OP.REPLACE is not allow above api 26

lvfen:
fix issue#1720

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/utils/BoxShadowUtil.java
+++ b/android/sdk/src/main/java/com/taobao/weex/utils/BoxShadowUtil.java
@@ -399,7 +399,13 @@ public class BoxShadowUtil {
       Rect newRect = new Rect(bounds);
       // Make the Canvas Rect bigger according to the padding.
       newRect.inset(-paddingX * 2, -paddingY * 2);
-      canvas.clipRect(newRect, Region.Op.REPLACE);
+
+      // Region.Op.REPLACE is not allow above api 26
+      if (Build.VERSION.SDK_INT >= 26) {
+        canvas.clipRect(newRect);
+      } else {
+        canvas.clipRect(newRect, Region.Op.REPLACE);
+      }
 
       Path contentPath = new Path();
       // the content area map must be aligned with bounds


android/sdk/src/main/java/com/taobao/weex/utils/BoxShadowUtil.java
```diff
@@ -399,7 +399,13 @@ public void draw(Canvas canvas) {
       Rect newRect = new Rect(bounds);
       // Make the Canvas Rect bigger according to the padding.
       newRect.inset(-paddingX * 2, -paddingY * 2);
-      canvas.clipRect(newRect, Region.Op.REPLACE);
+
+      // Region.Op.REPLACE is not allow above api 26
+      if (Build.VERSION.SDK_INT >= 26) {
+        canvas.clipRect(newRect);
```
YorkShen: `canvas.clipRect` is the same as `canvas.clipRect(newRect, Region.Op. INTERSECT)`, which is different from `canvas.clipRect(newRect, Region.Op.REPLACE)`.

lvfen: It is different, but the problem is that `Region.Op.REPLACE` is invalid after Android P. In order to avoid crash, use `Region.Op.INTERSECT` to replace `Region.Op.REPLACE`.

lvfen: Here is the android source code:
`private static void checkValidClipOp(@NonNull Region.Op op) {
        if (sCompatiblityVersion >= Build.VERSION_CODES.P
                && op != Region.Op.INTERSECT && op != Region.Op.DIFFERENCE) {
            throw new IllegalArgumentException(
                    "Invalid Region.Op - only INTERSECT and DIFFERENCE are allowed");
        }
    }

    /**
     * Modify the current clip with the specified rectangle.
     *
     * @param rect The rect to intersect with the current clip
     * @param op How the clip is modified
     * @return true if the resulting clip is non-empty
     *
     * @deprecated Region.Op values other than {@link Region.Op#INTERSECT} and
     * {@link Region.Op#DIFFERENCE} have the ability to expand the clip. The canvas clipping APIs
     * are intended to only expand the clip as a result of a restore operation. This enables a view
     * parent to clip a canvas to clearly define the maximal drawing area of its children. The
     * recommended alternative calls are {@link #clipRect(RectF)} and {@link #clipOutRect(RectF)};
     *
     * As of API Level API level {@value Build.VERSION_CODES#P} only {@link Region.Op#INTERSECT} and
     * {@link Region.Op#DIFFERENCE} are valid Region.Op parameters.
     */
    @Deprecated
    public boolean clipRect(@NonNull RectF rect, @NonNull Region.Op op) {
        checkValidClipOp(op);
        return nClipRect(mNativeCanvasWrapper, rect.left, rect.top, rect.right, rect.bottom,
                op.nativeInt);
    }`

lvfen: Here is the android source code:

`private static void checkValidClipOp(@NonNull Region.Op op) {
        if (sCompatiblityVersion >= Build.VERSION_CODES.P
                && op != Region.Op.INTERSECT && op != Region.Op.DIFFERENCE) {
            throw new IllegalArgumentException(
                    "Invalid Region.Op - only INTERSECT and DIFFERENCE are allowed");
        }
    }`

`/**
     * Modify the current clip with the specified rectangle.
     *
     * @param rect The rect to intersect with the current clip
     * @param op How the clip is modified
     * @return true if the resulting clip is non-empty
     *
     * @deprecated Region.Op values other than {@link Region.Op#INTERSECT} and
     * {@link Region.Op#DIFFERENCE} have the ability to expand the clip. The canvas clipping APIs
     * are intended to only expand the clip as a result of a restore operation. This enables a view
     * parent to clip a canvas to clearly define the maximal drawing area of its children. The
     * recommended alternative calls are {@link #clipRect(RectF)} and {@link #clipOutRect(RectF)};
     *
     * As of API Level API level {@value Build.VERSION_CODES#P} only {@link Region.Op#INTERSECT} and
     * {@link Region.Op#DIFFERENCE} are valid Region.Op parameters.
     */
    @Deprecated
    public boolean clipRect(@NonNull RectF rect, @NonNull Region.Op op) {
        checkValidClipOp(op);
        return nClipRect(mNativeCanvasWrapper, rect.left, rect.top, rect.right, rect.bottom,
                op.nativeInt);
    }`

YorkShen: After some research, I found the implementation of box shadow relies on `canvas.clipRect(newRect, Region.Op.REPLACE);`. Without this line, the box-shadow will never draw beyond the view's bounds.

I am going to disable box-shadow on Android P or higher and merged your PR.

Thanks

YorkShen:
Closed by #2545

PR #2321, opened 2019-04-15
[core] fix flex-end in row-reverse

win80540:
We fount that, when flex-direction is row-reverse, justify-content:flex-end is not available, not matter justify-content is flex-start or flex-end, layout system always layout items as flex-start.

Demo:
https://jsplayground.taobao.org/raxplayground/a3ad1d37-a6b7-41bb-9b72-3c00ab4ee161

### DIFF
--- a/weex_core/Source/core/layout/layout.cpp
+++ b/weex_core/Source/core/layout/layout.cpp
@@ -715,7 +715,9 @@ namespace WeexCore {
 
     for (WXCoreFlexLine *flexLine: lines) {
       float spaceBetweenItem = 0.f;
-      layoutFlexlineHorizontal(width, flexLine, childLeft, childRight, spaceBetweenItem);
+      // Bugfix: when flex-direction is row-reverse, justify-content:flex-end is not available, it always performance as flex-start
+      // layoutFlexlineHorizontal(width, flexLine, childLeft, childRight, spaceBetweenItem);
+      layoutFlexlineHorizontal(isRtl, width, flexLine, childLeft, childRight, spaceBetweenItem);
       spaceBetweenItem = std::max(spaceBetweenItem, 0.f);
 
       if(absoulteItem == nullptr) {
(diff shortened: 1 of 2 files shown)


gubaojian:
@YorkShen  layout preview

weex_core/Source/core/layout/layout.cpp
```diff
@@ -740,46 +742,66 @@ namespace WeexCore {
...
+              denominator = visibleItem != 1 ? visibleItem - 1 : 1.f;
+              spaceBetweenItem =
+              (width - flexLine->mMainSize - sumPaddingBorderAlongAxis(this, true)) / denominator;
+              childRight = width - getPaddingRight() - getBorderWidthRight();
+              break;
+          case kJustifyFlexStart:
+          default:
+              if (isRTL) {
+                  childLeft = width - flexLine->mMainSize - getPaddingRight() - getBorderWidthRight();
+                  childRight = width - getPaddingRight() - getBorderWidthRight();
```
YorkShen: I think this line should change to  `childRight = width - getPaddingLeft() - getBorderWidthLeft();`,  which is the same as kJustifyFlexEnd case with LTR.

The reason `childRight = width - getPaddingLeft() - getBorderWidthLeft();` works is due to the following equation:

    width = flexLine->mMainSize + getPaddingLeft() + getBorderWidthLeft() + getPaddingRight() + getBorderWidthRight()

Then we have

    childRight = getPaddingRight() + getBorderWidthRight()+ flexLine->mMainSize  =  width - getPaddingLeft() - getBorderWidthLeft();

win80540: Is not 
```
childRight = getPaddingLeft() + getBorderWidthLeft()+ flexLine->mMainSize  =  width - getPaddingRight() - getBorderWidthRight();
```
?
childRight contain `paddingRight` and `BorderWidthRight`?

YorkShen: I think your explanation is right. And we could have more brief explanation:
    
    childLeft = width - flexLine->mMainSize - getPaddingRight() - getBorderWidthRight();
    childLeft + flexLine->mMainSize  =  width - getPaddingRight() - getBorderWidthRight();
    childRight = width - getPaddingRight() - getBorderWidthRight();
    

As `kJustifyFlexStart` plus `RTL` equals `kJustifyFlexEnd` plus `LTR`, in which case, you may need to fix `kJustifyFlexEnd` plus `LTR`

PR #2337, opened 2019-04-19
[Android] remove unnecessary frameworkInit check in WXBridgeManager for data-render mode

hpop1994:
https://github.com/apache/incubator-weex/projects/3

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/WXSDKInstance.java
+++ b/android/sdk/src/main/java/com/taobao/weex/WXSDKInstance.java
@@ -2058,7 +2058,7 @@ public class WXSDKInstance implements IWXActivityStateListener,View.OnLayoutChan
     if (adapter == null) {
       return false;
     }
-    String result = adapter.getConfig("wxapm", "disable_skip_framework_init", "false");
+    String result = adapter.getConfig("wxeagle", "disable_skip_framework_init", "false");
     return "true".equals(result);
   }
 }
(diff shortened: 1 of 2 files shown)


android/sdk/src/main/java/com/taobao/weex/bridge/WXBridgeManager.java
```diff
@@ -2368,7 +2382,7 @@ public void reportJSException(String instanceId, String function,
       exception +=   "\n getTemplateInfo==" +instance.getTemplateInfo();//add network header info
       if (METHOD_CREATE_INSTANCE.equals(function) || !instance.isContentMd5Match()) {
         try {
-          if (isJSFrameworkInit() && (reInitCount > 1 && reInitCount < 10) && !instance.isNeedReLoad()) {
+          if (!isSkipFrameworkInit(instanceId) && isJSFrameworkInit() && (reInitCount > 1 && reInitCount < 10) && !instance.isNeedReLoad()) {
```
YorkShen: I think this line is wrong, consider changing to 

    ( isSkipFrameworkInit(instanceId) || isJSFrameworkInit() ) && (reInitCount > 1 && reInitCount < 10) && !instance.isNeedReLoad()

android/sdk/src/main/java/com/taobao/weex/WXSDKInstance.java
```diff
@@ -716,7 +719,7 @@ private void renderInternal(String pageName,
 
     final IWXJscProcessManager wxJscProcessManager = WXSDKManager.getInstance().getWXJscProcessManager();
 
-    if(wxJscProcessManager != null && wxJscProcessManager.shouldReboot()) {
+    if(wxJscProcessManager != null && wxJscProcessManager.shouldReboot() && !skipFrameworkInit()) {
```
YorkShen: I don't think you need changing this line.

The `if` case you changing is about a jscReboot mechanism, which will reboot JSC if **Weex Render** doesn't finished in the given `rebootTimeout()`.  The reboot mechanism mainly used to prevent deadlock of JSC, which has no relationship with dataRender.

PR #2424, opened 2019-05-09
[Android] feature : check bundle_type and report error

lucky-chen:
check bundle_type and report error

iOS pr:https://github.com/apache/incubator-weex/pull/2420/files

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/WXHttpListener.java
+++ b/android/sdk/src/main/java/com/taobao/weex/WXHttpListener.java
@@ -89,6 +89,10 @@ public class WXHttpListener implements IWXHttpAdapter.OnHttpListener {
         this.instance = instance;
     }
 
+    protected WXSDKInstance getInstance() {
+        return instance;
+    }
+
     @Override
     public void onHttpStart() {
         if (this.instance != null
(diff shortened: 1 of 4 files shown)


android/sdk/src/main/java/com/taobao/weex/bridge/WXBridgeManager.java
```diff
@@ -1602,6 +1602,16 @@ private void invokeCreateInstance(@NonNull WXSDKInstance instance, Script templa
           }
           return;
         } else {
+          if (instance.getRenderStrategy() != WXRenderStrategy.DATA_RENDER_BINARY && instance.getRenderStrategy() != WXRenderStrategy.DATA_RENDER) {
```
YorkShen: The `if` in 1592 already considered DATA_RENDER or DATA_RENDE_BINARY.

No need to do it twice.

android/sdk/src/main/java/com/taobao/weex/WXHttpListener.java
```diff
@@ -89,6 +89,10 @@ public void setSDKInstance(WXSDKInstance instance) {
         this.instance = instance;
     }
 
+    public WXSDKInstance getInstance() {
```
YorkShen: `protceted` will be better.

PR #2536, opened 2019-06-12
[Android] fix Android9 inaccurate ScreenHeight problem

katherine95s:


android/sdk/src/main/java/com/taobao/weex/utils/WXViewUtils.java
```diff
@@ -207,7 +210,14 @@ public static int getScreenHeight() {
   public static int getScreenHeight(Context cxt) {
     if(cxt!=null){
       Resources res = cxt.getResources();
-      mScreenHeight =cxt.getResources().getDisplayMetrics().heightPixels;
+      if(Build.VERSION.SDK_INT >= 28){
+        Point size = new Point();
+        ((WindowManager)cxt.getSystemService(Context.WINDOW_SERVICE)).getDefaultDisplay().getRealSize(size);
```
YorkShen: `((WindowManager)cxt.getSystemService(Context.WINDOW_SERVICE)).getDefaultDisplay()` may be null, add an `if` case or `try catch` when something unexpected happens then return the old-style screenHeight `cxt.getResources().getDisplayMetrics().heightPixels;` as before.

PR #2561, opened 2019-06-17
Fork/support arm64

Darin726:
Make weex Support arm64-v8a


android/sdk/build.gradle
```diff
@@ -114,16 +127,32 @@ android {
 //        include 'weex-rax-extra-api.js'
 //    }
 
+    if(project.hasProperty('asfRelease')){
```
YorkShen: The `if(project.hasProperty('asfRelease'))` block is not needed anymore, you can delete it.

android/sdk/build.gradle
```diff
@@ -79,9 +87,14 @@ if (!project.hasProperty('ignoreVersionCheck') || !project.getProperty('ignoreVe
 
 android {
     compileSdkVersion project.compileSdkVersion
-    buildToolsVersion project.buildToolsVersion
+//    buildToolsVersion project.buildToolsVersion
     resourcePrefix "weex"
     useLibrary 'org.apache.http.legacy'
+    packagingOptions {
```
YorkShen: In order to make Android App with or without `libc++_shared.so` run as expected, you need to do the following things:
1. delete `libc++_shared.so` in `libs` directory
1. Add a command flag `./gradlew -PIgnoreLibC=true` to enable `packagingOptions`. The `packagingOptions` should be disabled as default.

android/playground/app/src/main/AndroidManifest.xml
```diff
@@ -128,15 +128,15 @@ under the License.
...
-        <provider
-            android:name="android.support.v4.content.FileProvider"
-            android:authorities="com.alibaba.weex.fileprovider"
-            android:grantUriPermissions="true"
-            android:exported="false">
-            <meta-data
-                android:name="android.support.FILE_PROVIDER_PATHS"
-                android:resource="@xml/filepaths" />
-        </provider>
+        <!--<provider-->
```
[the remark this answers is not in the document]
Darin726: this will make playground crash

wuhh2016:
什么时候发布？

PR #2588, opened 2019-06-20
Fix iconFont render error when download from network first time

zshshr:
Fix iconFont render error when download from network first time


zshshr:
https://jsplayground.taobao.org/raxplayground/fc7a8d29-6e98-48bc-b5bb-a7d9a25a2e16

PR #2606, opened 2019-06-24
[iOS] fix bundleType

myeveryheart:
bundleType can be Vue, Rax
when the bundleType is "vue", the program goes bundleType = "other"


myeveryheart:
@wqyfavor Please review, the fix is very important to our product!

YorkShen:
I am not an iOS developers, but at least you should support `Vue` and `vue` together.

myeveryheart:
As you wish, I support Vue/vue and Rax/rax together.

PR #2638, opened 2019-07-02
[Android] bugfix of  the scrollstart and scrollend  event mismatching

zshshr:
[Android] bugfix of  the scrollstart and scrollend  event mismatching
* [water-fall](https://alibaba.github.io/rax/jsplayground/f0e48f35-b8d0-4365-9a77-561c9b72db24)
* [list-view](https://alibaba.github.io/rax/jsplayground/d05c8037-007f-4c3e-b5a3-4d460b45f3ee)
* [recycler-view](https://alibaba.github.io/rax/jsplayground/2f808e14-3f1a-452f-ac95-1fe7592ac876)

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/ui/component/helper/ScrollStartEndHelper.java
+++ b/android/sdk/src/main/java/com/taobao/weex/ui/component/helper/ScrollStartEndHelper.java
@@ -42,7 +42,10 @@ public class ScrollStartEndHelper implements Runnable{
     private Handler handler;
     private WXComponent component;
     private boolean  hasStart;
+    private boolean  canStart = false;
+
     private long  minInterval;
+    private int oldState = OnWXScrollListener.IDLE;
 
     private int x;
     private int y;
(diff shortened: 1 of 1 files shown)


android/sdk/src/main/java/com/taobao/weex/ui/component/helper/ScrollStartEndHelper.java
```diff
@@ -64,14 +67,16 @@ public void  onScrolled(int x, int y){
                 || component.getEvents().contains(Constants.Event.SCROLL_END))){
             this.x = x;
             this.y = y;
-            if(!hasStart){
+            if(!hasStart && canStart){
```
YorkShen: This `if` block should be as code in line 95  fired the event.

zshshr: code inline 95 is for the case the list has reached the top or bottom that can not callback in onscroll

YorkShen: Actually, you can safely delete code from line 95 to line 98,  which are never invoked in any case.

PR #2681, opened 2019-07-11
TestTravis

katherine95s:


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "weex-playground"]
+	path = weex-playground
+	url = https://github.com/katherine95s/incubator-weex.git
```
YorkShen: You should change this to `https://github.com/apache/incubator-weex-playground`

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "weex-playground"]
+	path = weex-playground
```
YorkShen: Add weex-playground to `.gitignore.`

android/build.gradle
```diff
@@ -61,6 +61,19 @@ subprojects {
         //Default value for disableCov is false
         disableCov = project.hasProperty("disableCov") && disableCov.equals("true")
     }
+    if(project.name == 'playground'){
```
YorkShen: You should add a script to upgrade submodule automatically.

PR #2712, opened 2019-07-16
Update for libweexjsb path

Darin726:
1. add DefaultSetting SharedPreferences
2. unify so copy path with cpu type

### DIFF
--- a/android/sdk/src/main/java/com/taobao/weex/utils/WXSoInstallMgrSdk.java
+++ b/android/sdk/src/main/java/com/taobao/weex/utils/WXSoInstallMgrSdk.java
@@ -76,10 +76,11 @@ public class WXSoInstallMgrSdk {
   private final static String STARTUPSO = "/libweexjsb.so";
   private final static String STARTUPSOANDROID15 = "/libweexjst.so";
 
-
   static Context mContext = null;
   private static IWXSoLoaderAdapter mSoLoader = null;
   private static IWXStatisticsListener mStatisticsListener = null;
+
+  private static String mAbi = null;
 
   public static void init(Context c,
                           IWXSoLoaderAdapter loader,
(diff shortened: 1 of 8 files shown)


android/sdk/src/main/java/com/taobao/weex/utils/WXSoInstallMgrSdk.java
```diff
@@ -381,18 +351,19 @@ private static String _getFieldReflectively(Build build, String fieldName) {
...
-    try {
-      abi = Build.CPU_ABI;
-    }catch (Exception e){
-      e.printStackTrace();
-      abi = ARMEABI;
-    }
-    if (TextUtils.isEmpty(abi)){
-      abi = ARMEABI;
+    if(TextUtils.isEmpty(mAbi)) {
+      try {
```
YorkShen: 1. No need to add try catch. Even if there is `NoSuchFieldError`, it is an error, not an exception, so `catch Exception` is useless.
1. Use `Build.SUPPORTED_ABIS` as `Build.CPU_ABI` is deprecated.

android/sdk/src/main/java/com/taobao/weex/utils/WXSoInstallMgrSdk.java
```diff
@@ -208,110 +211,83 @@ public static boolean initSo(String libName, int version, IWXUserTrackAdapter ut
...
-            w.close();
         }
+      }
+
+      if(!oldfile.exists()) {
+        WXEnvironment.extractSo();
+        oldfile = new File(copyPath, STARTUPSO);
+      }
 
+      if (oldfile.exists()) {
```
YorkShen: The old code for line 287 is a `else` case. You are changing the behaviour in new code.

PR #2713, opened 2019-07-16
update TravisCI  for  iOS.

wrmswindmill:
the test case have been not updated for a long time, so I  comment the codes which are error.


.travis.yml
```diff
@@ -6,6 +6,8 @@ matrix:
     include:
       - env: TEST_SUITE=danger 
       - env: TEST_SUITE=jsfm
+      - env: TEST_SUITE=ios
```
YorkShen: No need to add `os: osx`?

wrmswindmill: image

Reference link:https://www.objc.io/issues/6-build-tools/travis-ci/

PR #2715, opened 2019-07-16
[android] 修复eagle在Framework尚未init时遇到的platformBridge缺失的问题

hpop1994:


weex_core/Source/android/utils/jni_load.cc
```diff
@@ -51,7 +54,8 @@ jint JNI_OnLoad(JavaVM *vm, void *reserved) {
     WeexCore::SoUtils::Init(env);
     WeexCore::WMLBridge::RegisterJNIUtils(env);
   }
-
+  PlatformBridge* bridge = new AndroidBridgeInSimple;
+  WeexCoreManager::Instance()->set_platform_bridge(bridge);
```
YorkShen: Remove the following lines in wx_bridge.cpp :

https://github.com/apache/incubator-weex/blob/67bad4f44cd6aaabd46d532c2050228d0f669e5b/weex_core/Source/android/wrap/wx_bridge.cpp#L302-L303

PR #2899, opened 2019-09-16
add For Webview get OnActivity Result

Darin726:
add For Webview get OnActivity Result


android/sdk/src/main/java/com/taobao/weex/WXSDKInstance.java
```diff
@@ -203,6 +203,8 @@
 
   private List<OnBackPressedHandler> mWXBackPressedHandlers;
 
+  private List<OnActivityResultHandler> mWXOnActivityResultHandlers;
```
YorkShen: Change it to `Set`, which will gives you better performance.

PR #2783, opened 2019-08-02
[Android]support "wx" in animation and wxtext

katherine95s:

PR #2923, opened 2019-09-19
Bugfx/revert 2783 2883

lucky-chen:
- revert #2783 
- revert #2883


YorkShen:
Revert documentation as well.

https://github.com/apache/incubator-weex-site/pull/491

PR #2940, opened 2019-09-26
* [Android] Remove shared library from binary of Weex

YorkShen:
* Use [jsc-android](https://www.npmjs.com/package/jsc-android) under BSD Licnese to replace LGPL dependency. The jsc-android dependency only exists in compiling time, there is no limited in runtime for users, they could use whatever they like.
* Remove LGPL files(`*.so`) in the convenience binary of Weex, which will give users the freedom of choosing JS interpreter at runtime.


YorkShen:
En...

It seems like the `URL` obejcet in groovy doesn't support redirection, which cause Travis Failure. I will look into it later.

YorkShen:
The URL redireciton is fixed by using to https://github.com/apache/incubator-weex/blob/release/0.26/android/sdk/weex_sdk-0.26.0.aar.

Issue #2971, opened 2019-10-17
[iOS]jsMainBundleStringContentMd5写入错误


jqoo:
` 

   if (!self.userInfo[@"**jsMainBundleStringContentLength**"]) {
        self.userInfo[@"jsMainBundleStringContentLength"] = @([mainBundleString length]);
    }
    if (!self.userInfo[@"**jsMainBundleStringContentLength**"]) {
        self.userInfo[@"jsMainBundleStringContentMd5"] = [WXUtility md5:mainBundleString];
    }
`
判断条件都是写的jsMainBundleStringContentLength，第二个key应该是jsMainBundleStringContentMd5才对

PR #2994, opened 2019-10-31
[Fix ]md5 value of bundle string was set for a wrong key

jqoo:
_renderWithMainBundleString中填写md5值时使用了错误的key

Fix issue: [#2971](https://github.com/apache/incubator-weex/issues/2971) 

# Brief Description of the PR

# Checklist
* Demo:
* Documentation:


wqyfavor:
I think we should remove the logic of storing "jsMainBundleStringContentMd5" in user info.
Thanks anyway,

PR #3026, opened 2019-11-20
[Android] fix memleak when stop performance record

lucky-chen:
fix apache_master_new when stop apache_master_new performance data


YorkShen:
Duplicated with https://github.com/apache/incubator-weex/pull/3031

PR #3027, opened 2019-11-20
[Android] fix weexcore mmap leak

lucky-chen:
fix weexcore mmap leak


YorkShen:
Closed due to duplicated with https://github.com/apache/incubator-weex/pull/3044

